Issue #110, opened 2026-09-16
Certificate button does nothing on the Android tablet; keep certificates like stickers under "My certificates" (view, zoom, download)


ugurozsahin:
Owner (playtest on the Android tablet, 2026-09-11): on the results screen the **certificate button can be pressed but nothing happens** — no share sheet, no save, no picture. On the phone and in the browser the same button works.

## Why, as far as the code shows

`src/ui/certificate.ts` picks a delivery route in `certRoute()`: Web Share with files → the artifact viewer's `downloads` save → a full-screen press-and-hold view → a plain `<a download>`. In the Capacitor WebView on the tablet there is no artifact runtime and `navigator.canShare({ files })` is false or `share` is absent, so the route falls through to `<a download>` (`triggerDownload`) — and a stock Android WebView has **no download handler**, so `a.click()` is swallowed silently. The function's own contract ("it never silently does nothing") is broken exactly on the device the APK exists for. Confirm this first on the tablet with `adb logcat` or a `toast` of the chosen route; the fix must not be guessed.

There is a second gap underneath: **certificates are not kept anywhere.** `drawCertificate(cert)` renders from the result data on the spot, and once the results overlay closes the certificate is gone. A child who could not save it has no way back to it.

## Owner's design — the sticker approach

Treat certificates like stickers: every certificate earned is **stored** and **listed under "My certificates"** (Ninja Rewards, next to the sticker album — `src/ui/home.ts` rewards screen). From the list a child or grown-up can:
- **open** one full-screen (the existing `.cert-view` view is the seed),
- **zoom** (pinch or tap-to-zoom on the image),
- **download / share** it — with a route that actually works on Android.

## Acceptance criteria (draft — owner may adjust)

- **Storage**: a certificate is saved as *data* (name, year, mode/topic, stars, score, date — whatever `drawCertificate` needs), not as a PNG, and redrawn on demand; `SAVE_VERSION` bump + `migrate()` per `storage.ts`, and the export/import save code (#64's #172) carries it. A cap or dedupe rule (one per topic-mode-stage, newest wins?) — propose in the PR.
- **List**: "My certificates" on the rewards screen, thumbnails, newest first; empty state that tells the child how to earn one. The results screen keeps its button, which now also files the certificate into the list.
- **View**: full-screen with zoom; Done/back; keyboard and tap-outside close. Uses `screenScope()`/existing overlay patterns, no new abstraction.
[...]

Issue #96, opened 2026-09-16
CI: run e2e only when the change can affect the game


ugurozsahin:
Owner (session, 2026-09-10): the e2e suite runs on changes that cannot possibly affect the game, and he asked whether it can be made to trigger only on real game changes — and whether the same applies to the unit tests.

## What actually happens today
`.github/workflows/ci.yml` has `paths-ignore: ['**.md', 'docs/**', '.claude/**', '.gitignore']`, so a **pure** docs change is correctly skipped. The problem is that the filter is **workflow-level and all-or-nothing**, and the e2e step has no filter of its own — so one non-doc file dragging along takes the whole 9-minute suite with it. Measured on today's runs:

| head | changed files | e2e |
|---|---|---|
| `df78a88` (PR #169) | `BACKLOG.md`, `CLAUDE.md`, `docs/ROUTINE-PROMPT.md`, `docs/WATCHDOG-PROMPT.md`, `tests/unit/guardrails.test.ts` | **ran** |
| `1f00336` (PR #167) | `tests/unit/guardrails.test.ts` only | **ran** |

Neither could change a pixel of the game. Governance PRs hit this every time, because the rails that enforce the docs live in `tests/unit/`, so a docs change is almost never docs-only.

## What to build
Keep the workflow running on everything, and make **the e2e step** conditional on the change actually touching the game:

- **always:** unit tests + guard rails (14 s) and `npm run build` (which is where `tsc --noEmit` lives).
- **e2e only when** the diff touches `src/**`, `index.html`, `public/**`, `tests/e2e/**`, `playwright.config.*`, `package.json` / `package-lock.json`, or `.github/workflows/ci.yml`.

**Do not implement this with a workflow-level `paths:` filter.** That makes the whole workflow not run, and a PR then has *no* `CI` check at all — which collides head-on with the merge rule that "anything that touches code and has no run is a red light" and with reading a missing check as a pass. The workflow must still run and report; only the e2e step is skipped, visibly, so a reviewer can see it was skipped and why. A `git diff --name-only` step setting an output is enough — no new action dependency needed.

## The unit tests stay on everything, deliberately
[...]

Issue #70, opened 2026-09-16
Sticker album is too easy: unlock stickers from achievements, not lifetime coins


ugurozsahin:
Owner (2026-09-07): the sticker album in Ninja Rewards is far too easy to complete. **Owner's decision: move the album from "lifetime coins" to achievements** - the first few stickers stay on a cheap coin threshold, the rest unlock from things the child has actually done.

## The maths today
`src/storage.ts:102` - `STICKER_COST = [30, 70, 120, 180, 250, 330, 420, 520, 630, 750, 900]`, and `stickersFor(coins)` unlocks purely from **lifetime** coins.

Earning (`src/game/modes.ts:47,57`): `baseCoins = correct + stageStarsTotal * 5`, plus 20 for winning a mission. A Reception mission is 5 stages x 5 questions (`src/curriculum/types.ts:71`), so a good run is roughly 23 correct + 15 stage stars x 5 = 75, + 20 = **~120 coins per mission**. The last sticker sits at 900 lifetime coins, so the **whole album is ~8-10 missions - a single afternoon.**

Worse than the number: coins are a pure volume metric. A child can replay one easy topic and unlock everything without ever touching a second topic, a boss, or a harder year. The album measures time spent, not what was learned.

## What to build
- **Keep the first 2-3 stickers on coins** at their current cheap thresholds (30 / 70 / 120). A four- or five-year-old needs a visible win in the first session; do not take that away.
- **The remaining 8 unlock from achievements** - deeds, not grinding. Candidate conditions (final list is the developer's proposal, owner's approval):
  - star N different topics (breadth across the curriculum - the pedagogically valuable one)
  - star at least one topic on each island / year the child has access to
  - win a Boss Battle; win one without losing a life
  - complete a Memory Match board
  - a 3-day and a 7-day streak
  - clear a mission stage with no wrong answers
  - a Ninja Sprint personal best above a threshold
  - complete every topic in one island
- Conditions must be **derivable from data already in the save** (`progress`, `endless`, `sprint`, `boss`, `memory`, `training`, `streak`, `dojo` in `src/storage.ts`) wherever possible; add new counters only where there is no other way, and put them through `migrate()` (guard rail #38).
- **Nothing is ever taken away.** A sticker already earned stays earned after this change - if a current player's album shows 11/11 from coins, it must not empty out on update. Decide and document the migration: simplest is to keep every sticker the save already lists as unlocked, and apply the new rules only to ones not yet earned.
[...]

Issue #69, opened 2026-09-16
Ninja Shop: sell every element slice trail (colour + effect), price-laddered


ugurozsahin:
**Blocked by #114** (owner, session 2026-09-11): both touch the reward economy; the album moves to achievements first, then the trail price ladder is set against it. The routine skips this until #114 closes; the `blocked` label puts it in the board's Blocked column and comes off when the blocker lands.

Owner (2026-09-07): every ninja's slice trail should be buyable in the Ninja Shop - so a child playing as Frost can save up for the fire trail. **Owner's decision: a bought trail changes the colour *and* the particle effect, not just the colour.**

## Why
The reward loop is thin: `SHOP_ITEMS` (`src/game/shop.ts:11`) has exactly two trails - the free `trail-element` and the 200-coin `trail-gold`. Once a child has the golden trail there is nothing left to want, which is the "nothing to spend coins on" problem the owner raised. Meanwhile all ten element effects already exist in `src/game/arena.ts` (`FxKind`, `FX_PARTICLE`, `FX_COLORS`) - so this is a catalogue change, not new artwork or new draw code, and it is the cheapest large addition available to the shop.

## What to build
- One shop item per element trail: fire, water, electric, earth, wind, ice, light, shadow, blade, robot. `kind: 'trail'`, each with the element's colours from `FX_COLORS` and a matching icon/blurb in British English.
- **Effect, not just colour.** `ShopItem` currently carries only `trail?: { color, core }`, while the particle kind comes from the *avatar*: `src/ui/play.ts:143` passes `fx: av.fx`. Add an `fx?: FxKind` to `ShopItem` and let an equipped trail override `av.fx`; falling back to the avatar's own element when no trail is equipped. Recolouring alone was rejected - a bought "fire trail" that still puffs shadow smoke reads as cheap.
- **`master` is not for sale.** The Master Ninja's effect is the reward for starring every topic (`masterProgress()`); selling it would undercut that unlock. Keep it out of the catalogue.
- **A price ladder, not a flat price.** Coins are earned at roughly `baseCoins + 20` per won mission (`src/game/modes.ts:57-84`) and stickers already unlock at 30/70/120/.../900 lifetime coins. A five-year-old needs a first purchase inside the first session or two, so start the cheapest trail low (~40-60) and climb, with `trail-gold` (200) staying the premium item at the top. Exact numbers are the developer's proposal, the owner's call.
[...]

PR #158, opened 2026-09-16
migrate() down-stamps a save from a newer build, and the next save() makes it permanent (#232)

### LINKED ISSUE #127: migrate() down-stamps a save from a newer build, and the next save() makes it permanent
Raised while reviewing PR #230 (by the reviewing run), confirmed by its author, and deliberately kept out of that PR to avoid widening it.

## The fault

`migrate()` ends:

```ts
while (v < SAVE_VERSION && MIGRATIONS[v]) { s = MIGRATIONS[v](s); v++; }
return { ...DEFAULT, ...s, v: SAVE_VERSION };
```

A blob from a **newer** build — `{ v: 3, … }` — fails the loop condition immediately (`3 < 2` is false) and falls straight through to the return, which stamps `v: 2` over v3-shaped data. `load()` caches that, and the next `save()` writes it back. The newer save is now permanently mislabelled as the older shape, with whatever v3 added still sitting in it unrecognised, and whatever v3 *renamed* silently read under its old name.

`importSave()` already refuses this case, with the reasoning spelled out in its docblock — *"A code from a newer build is refused too — migrations only run forwards, so there is nothing to bring a v2 save down to v1"* — and `storage.ts:213` enforces it. `load()` → `migrate()` is the same situation and does not apply the same rule.

## Why now

Entirely academic until this week: there was only ever v1, so "a blob from a newer build" could not exist. #230 (certificate album) and #222 (the TTS verdict) each take the save to v2, so a second version is about to exist for the first time, and the sequence that produces the fault is ordinary:

- the child's tablet gets a build with v3 while the phone is still on v2 (the APK and the web build do not update together — #64 is the whole reason the save code exists);
- a save code exported from the tablet, or a shared device rolled back, and `load()` down-stamps it.

The migration ladder is the thing keeping saves honest, and this is the one direction it does not defend.

## Proposed fix

`migrate()` cannot upgrade a future save and must not pretend to. Two candidate behaviours, and the choice is the substance of this issue:

1. **Refuse and keep the blob untouched** — return the default for this session but do not let `save()` overwrite the stored newer blob, so downgrading a build and upgrading again does not lose the child's progress. Needs `save()` to know it is holding a refused blob.
2. **Refuse and reset** — simplest, and matches `importSave`'s existing shape, but it silently discards a save the child's other device can still read.

(1) is the better answer for a child with two devices; (2) is two lines. Propose in the PR.

A guard rail either way: a `v` above `SAVE_VERSION` must never be written back as `SAVE_VERSION`.

## Related, same function, worth folding in

- **A non-numeric `v`** (`"2"`, `null`, `{}`) reads as a pre-versioning blob via `saveVersionOf`'s `typeof s.v === 'number'` test, so it re-runs the whole ladder over data that has already been migrated. Harmless for the additive v1 → v2 steps; not harmless for a reshaping one.
- **`save()` swallows the write failure** (`storage.ts:82-86`): `cache` is assigned before `setItem`, so a throwing write (private mode, DOM storage disabled in a WebView, quota) leaves the running app reporting state that is gone on the next launch. Pre-existing and consistent with the rest of the file, but #205's album is the first feature whose entire purpose is surviving a device that cannot save — a caller that could *know* the write failed is what a "your progress is not being saved" notice would need. Possibly its own issue; noted here so it is not lost.

## Acceptance criteria

- A stored blob with `v > SAVE_VERSION` is never rewritten as `SAVE_VERSION`; the chosen behaviour above is implemented and its reasoning is in the docblock.
- A non-numeric `v` does not re-run the ladder.
- Unit tests for both, each mutation-checked red.
- A guard rail for the down-stamp, since it is silent by nature.
- No budget number rises.

### Workflow (minimum scrum - the developer and the reviewer/QA must be different agents)
- [ ] **Refine** - acceptance criteria above; the behaviour choice is the open question
- [x] **Develop** - branch `fix/232-migrate-downstamp`, PR #264
- [ ] **Review** - a *different* agent
- [ ] **QA** - unit + mobile e2e
- [ ] **Done** - merged, issue closed with the commit hash

---

---
_Migrated from #232 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #127.

`migrate()` only ever ran forwards, and had no answer for a blob that was already ahead of it. A `{ v: 3, … }` save failed `v < SAVE_VERSION` immediately and fell straight through to `return { ...DEFAULT, ...s, v: SAVE_VERSION }`, which stamps **our** version over v3-shaped data. `load()` cached that and the next ordinary `save()` wrote it back, so the mislabelling became permanent: whatever the newer version added sat unrecognised, and whatever it *renamed* was read under its old name. `importSave()` has refused a newer code since it was written, for exactly this reason — this is `load()`'s half of the same rule.

Ordinary sequence, not an exotic one: the APK and the web build do not update together (#57 is the whole reason the save code exists), so one of a child's two devices is routinely a version ahead of the other.

## Behaviour chosen: refuse and keep the blob untouched

The issue offered two, and preferred this one. The session runs on defaults and `save()` writes **nothing**, so the stored blob survives byte-for-byte and downgrading a build then upgrading again does not cost the child their progress. *Refuse and reset* was the two-line alternative and is rejected here: it silently discards a save the other device can still read, which is the same loss with a tidier implementation.

A module-level `readOnly` latch is what lets `save()` know it is holding a refused blob. It clears on `reset()` (the blob is gone, so there is nothing to protect) and on a successful `importSave()` (the grown-up deliberately replaced it).

**One consequence a reviewer should weigh rather than take on trust:** while the latch is set the child's session genuinely does not persist, and nothing tells them so. For the newer-save case that is the point — the alternative is destroying a save the other device still reads. It is less obviously right for an unreadable `v`, where the blob is garbage no build can read and preserving it means nothing can be written until a reset. I took the same branch for both because *we could not read it, so we do not touch it* is one rule rather than two, and `?reset` / "Start again" clears it. `isReadOnlySave()` is exported as the hook a *"your progress is not being saved"* notice would need — that notice is the issue's own related item and is deliberately not in this PR. If you think the unreadable case should reset instead, say so and I will split the branch.

## The second fault, same function

A `v` that is *present* but not a whole number ≥ 1 — `"2"`, `null`, `{}`, `1.5`, `-3` — read as a pre-versioning blob through `saveVersionOf`'s `typeof s.v === 'number'` test, so the whole ladder re-ran over data that had already been migrated. Harmless while every step is additive; silently destructive the first time a step reshapes. Absent `v` still means v1 (that is #58's fix and it stays); a present-but-unreadable `v` now reads as `UNREADABLE_VERSION` and is refused the same way a future save is, rather than guessed at.

This changes one existing expectation — `saveVersionOf({ v: 'two' })` was pinned at `1` — and the test now pins it at `UNREADABLE_VERSION` with the reasoning inline. Its original point, *never read as current*, is pinned separately and still holds.

## Tests, and both runs

Behavioural, deliberately: a text rail could check `isMigratable` is still *called* and would stay green through any change that kept the identifier and broke the comparison — #110's lesson.

**Prove it red** — each mutation restores the original code, against the new tests:

| mutation | `storage.test.ts` |
| --- | --- |
| clean tree | **35 passed** |
| `migrate()`'s refusal removed (the original down-stamp) | **3 failed** |
| `save()`'s read-only latch ignored (the original unconditional write) | **2 failed** |
| the original `saveVersionOf` restored | **3 failed** |

And the guard rail on its own:

| | `guard rail: … newer build (#127)` |
| --- | --- |
| bug restored | **1 failed** |
| clean tree | **1 passed** |

## Guard rail

`a save from a newer build is never relabelled as this version (#127)`, in `guardrails.test.ts` beside the two existing rails on the same seam (#38's `load()` routes through `migrate()`, #110's every-version-has-a-step). It asserts the one-line invariant — a newer blob's data must not come through under our version stamp — plus that the guard does not overreach, so every readable version still migrates. The full behaviour (untouched blob, latch set, latch cleared) lives in `storage.test.ts`.

**No budget number moved in either direction.**

## What was deliberately not done

- **The `save()`-swallows-the-write-failure item** at the foot of the issue (`cache` is assigned before `setItem`, so a throwing write leaves the app reporting state that is gone next launch). The issue itself says "possibly its own issue"; it is a different fault with a different fix and a user-visible notice attached, and folding it in would widen this PR. Not filed yet — worth one.
- **No `SAVE_VERSION` bump.** This PR makes a bump *safe*; it does not make one.

## Why this issue, out of strict oldest-first

Saying it plainly so a reviewer can disagree. The query's strict answer after #64 (PR #255 open) and #69 (`blocked`) is #67, then #70. #67's own acceptance criteria require the onboarding flag to go through `migrate()`, which means taking `SAVE_VERSION` to 3 — and #127's "Why now" section says a second version existing is exactly what turns this bug from academic into live data loss. Landing #67's storage slice first would knowingly ship that hazard. The reasoning is recorded on both issues (#67, #127) rather than only here.

## Tests run

| check | result |
| --- | --- |
| `npm test` | **951 passed** (30 files) — `main` at `a1ddfb3` is 945 |
| `npm run build` | ok |
| e2e `--project=mobile` | **52/52 passed** (3.4m) |

Desktop not run and not warranted: this diff is storage logic with no viewport-sensitive surface. CI will run e2e on this one — the diff touches `src/`, so #96's filter does not apply.

https://claude.ai/code/session_013rN4PdH74Q1zm2m5vTEfZP

---
_Generated by [Claude Code](https://claude.ai/code/session_013rN4PdH74Q1zm2m5vTEfZP)_

---
_Migrated from #264 in the archived private repo (ugurozsahin/sky-academy-private-archive). History rewritten to remove a third-party reference.__


ugurozsahin:
**Review: approved.** Oldest of the 4 open PRs, reviewed first per STEP 2.

## Verified locally (branch `fix/232-migrate-downstamp` @ `796535f`)
| check | result |
| --- | --- |
| `npx vitest run` | **987 passed** (30 files) |
| CI (GitHub Actions, this head) | green — [run 35091322792](https://github.com/ugurozsahin/sky-academy/actions/runs/35091322792), review-gate: "No open review block" |

e2e not re-run locally in this pass (no viewport-sensitive surface per the PR's own note, and CI already ran mobile e2e green on this exact head).

## Read against the diff, CLAUDE.md and the acceptance criteria on #127
The read-only latch is correctly scoped to *future* saves only (an unreadable `v` resets instead, with the reasoning for the asymmetry spelled out in the docblock and covered by its own test) — this is the sharper of the two candidate designs #127 offered, and the harder one to get right. `exportSave()` and `importSave()` both close the two secondary holes the PR describes finding during its own review (export handing out a blank code under the latch; a failed import write prematurely clearing the latch) — both are pinned by dedicated tests. No `SAVE_VERSION` bump, consistent with "what was deliberately not done".

## Vendored agents (#99)
- **pr-test-analyzer**: coverage is solid and maps to the real behaviour surface; two non-blocking nice-to-haves (a direct assertion that `save()` while latched still updates the in-session value even though the disk write is skipped, and a couple of standalone `saveVersionOf` boundary assertions rather than only exercising them via `load()`).
- **silent-failure-hunter**: flagged `exportSave()`'s `catch` around the second `getItem(KEY)` call as a theoretical fallback to `JSON.stringify(load())` (the in-memory default) if that call throws. Weighed this myself: `readOnly` can only be `true` if the immediately-preceding `load()` call in the same synchronous function already read that exact key successfully, so a `getItem` throwing on the very next line is not a realistic failure mode here — not blocking. Also flagged `importSave()` returning `true` even when the write silently fails — this is pre-existing behaviour the PR explicitly declines to widen (tracked as issue 266 in the PR body), not a regression.
- **type-design-analyzer**: `UNREADABLE_VERSION = 0` is a sentinel folded into a plain `number` return rather than a discriminated union, so nothing at the type level stops a stray `v === 0` comparison from treating it as real. Correct and well-commented today; a `{kind:'ok'|'future'|'unreadable', v}` union would make it compiler-enforced. Not urgent, noted for a future touch of this seam.

None of the above is blocking. No guard-rail budget moved except the new rail's own addition, matching the PR's claim.

Squash-merging.

---

Issue #155, opened 2026-09-16
CI is down: every Actions job that needs a runner fails at startup with zero steps — nothing can merge


ugurozsahin:
**Every GitHub Actions job that needs a runner is failing at startup. Nothing can merge and the nightly will not run.**

Observed by the 2026-09-15T20:37Z routine run at `2026-09-16T02:51Z`–`02:55Z` while handling PR events for #264.

## The signature

Three jobs, across **two** workflows, on head `2f66564` of `fix/232-migrate-downstamp`:

| check | workflow | run | attempt 1 | attempt 2 |
| --- | --- | --- | --- | --- |
| `test` | CI | `35049581223` | failure | failure |
| `branch-name` | CI | `35049581223` | failure | failure |
| `gate` | Review gate | `35049581215` | failure | failure |

Every one of them:

- executed **zero steps** (`steps: 0` from `/actions/runs/<id>/jobs`);
- produced an **empty log** (`/actions/jobs/<id>/logs` returns nothing);
- has no check-run `output.title` and no `output.summary`;
- lived **2–4 seconds** (`02:52:34Z → 02:52:37Z`).

A job with zero steps never reached checkout, so nothing in any diff ran.

## It is not the pull request that surfaced it

- `branch-name` only greps the head ref against `feature|fix|chore/<issue>-<slug>`. The ref is unchanged and that job **passed twice on this same branch** (`e6c1879`, `630e252`).
- `git diff origin/main HEAD -- .github/` on that branch is **empty** — no workflow was touched.
- Its whole diff is three source/test files.
- That same diff, on the previous base, was **fully green** including the e2e step (run `35033906844`, head `630e252`).
- The tree is green locally: `tsc` clean, `npm test` 987 passed, `npm run build` ok.

## What is and is not still working

An `issue_comment`-triggered **Review gate** run at `02:54:06Z` completed as **`skipped`** rather than `failure`. That matters: Actions is still receiving events, evaluating workflows and dispatching them for this repository. A `skipped` job never allocates a runner, so it does **not** show runners are healthy — but it does narrow the fault to **runner allocation**, not the Actions service as a whole.

## Leading hypothesis, explicitly not confirmed

This looks like an **Actions spending/quota stop**. The repository is private with metered minutes, and #119 is open with the spending limit deliberately closed; a billing stop halts runner allocation while leaving event processing intact, which is exactly what is observed.

[...]

Issue #131, opened 2026-09-16
Sound Hunt's spoken-path test races a 4-second real-time voice probe, so it fails intermittently


ugurozsahin:
> **Corrected 2026-09-15 by the filer.** The original title was "Desktop e2e is red on main", and that was wrong — the evidence did not support it. A different session ran the same test on desktop twice and it **passed** both times (comment below), and identified the mechanism: `VOICE_START_MS = 4000` in `src/audio.ts` is real wall-clock time, deliberately not scaled by the suite's 4×, so the test races that timer to assert `🔊 Listen!` before the probe falls and the card repaints with the keywords. That is a flaky test, not a desktop-specific red, and it is not about viewports at all. The original report is kept below unedited because the comment answers it point by point. **Criterion 1 is therefore already settled; start at criterion 3.**

---

`main` is red on the **desktop** e2e project. Found on 2026-09-14 while running the full matrix for PR #239, and confirmed against `main` itself rather than assumed.

## What fails

```
[desktop] › tests/e2e/game.spec.ts › Sound Hunt: nothing to read on the card;
                                     the words appear only when read-aloud is off

Error: expect(locator).toHaveText(expected) failed
Locator:  locator('.prompt')
Expected: "🔊 Listen!"
Received: "ten · tap · top"
```

The card is taking #65's **no-voice** path — printing the three keywords — where the test expects the spoken path and a picture-free "🔊 Listen!" card. Over the 5-second poll the locator resolved three times, each to a different question's keywords (`pen · pig · pot`, `add · axe · apple`, `ten · tap · top`), so the screen is live and answering; it has simply decided read-aloud is unavailable.

## What is and is not established

- **It is `main`'s, not a pull request's.** Checked out `main` at `cb621c4` and ran that single test on the desktop project: fails identically. PR #239's diff (a Year 2 statistics topic) touches no audio, speech or writing code.
- **It is not simply "no voices in this container".** The same test **passes on the mobile project** on both trees. Whatever differs between the two projects is the lead. *(Superseded: it is a timing race, and the same container passes desktop too on another run.)*
- **CI has never run desktop against #65.** A pull request runs mobile only (#141) and the full matrix is the nightly. The last nightly was `2026-09-14T08:55:18Z` on head `34dafed` — **before** #65 merged as `858d2d1` — so this has not yet had a chance to show up in CI. Tonight's nightly is the first that will. *(This part still stands and is the reason it is worth fixing now rather than after it reds a nightly.)*
[...]

Issue #129, opened 2026-09-16
Android APK: setup-android's default `packages` still asks for the retired `tools`, so every APK build fails


ugurozsahin:
**Found by the owner, running the workflow by hand on 2026-09-14T22:33Z.** Every "Android APK" run fails at the SDK step; the APK cannot be built at all.

```
[command]/usr/local/lib/android/sdk/cmdline-tools/16.0/bin/sdkmanager tools
Warning: Failed to find package 'tools'
Error: The process '…/sdkmanager' failed with exit code 1
```

## Cause

`.github/workflows/android.yml` calls the action with no inputs:

```yaml
      - uses: android-actions/setup-android@v3
```

so the action's **default** applies — `packages: 'tools platform-tools'`. `tools` is the legacy "Android SDK Tools" package, which Google has retired from the SDK repository; `sdkmanager` therefore cannot resolve it and exits 1. Nothing in this repository changed: the remote package list did. The licence-acceptance step just above it succeeds, which is why the log looks healthy until the last two lines.

## Fix

Name the packages explicitly so the action's default can never apply again, and ask only for what exists:

```yaml
      - uses: android-actions/setup-android@v3
        with: { packages: 'platform-tools' }
```

The platform and build-tools do not need listing: AGP 8.13 (`android/build.gradle`) downloads the ones `android/variables.gradle` asks for (`compileSdkVersion = 36`) once the licences are accepted, which this step still does. Listing them here instead would be a second place to keep in step with `variables.gradle`.

## Guard rail

A unit rail over `.github/workflows/android.yml` (`tests/unit/guardrails.test.ts`), because this failure is invisible to CI — the APK workflow runs only on `workflow_dispatch`, a `v*` tag, or a PR touching the Android paths, so nothing tells us it is broken until someone wants an APK:

1. the `setup-android` step declares `packages:` **explicitly** — that is what stops a silent third-party default from deciding what we install; and
2. the list never names the retired `tools` package (word-boundary matched, so `platform-tools` is fine).

Neither would have caught Google's retirement in advance, and the rail says so: what it catches is the shape that made us vulnerable — an input we do not control, left to its default.

**Detection is the other half, and it is cheap**: the watchdog can read the last "Android APK" run's conclusion through the API (no Actions minutes) and report a failed or absent run, the way it already reads the nightly. Proposed here, to be done with the fix.

[...]

Issue #74, opened 2026-09-16
watchdog: Actions budget projected over both Free and Pro quotas


ugurozsahin:
**Update — 2026-09-08T16:21Z (routine watchdog run).** Still over both quotas; recomputed independently from all 193 workflow runs (created_at→updated_at, rounded up to the minute — not `get_workflow_run_usage`, which still reports 0 on this repo):

- 193 runs total since Actions history began (2026-09-06T10:47Z) — repo still has no runs before Sept 1, so both windows still cover the same history.
- Raw total: **792 wall-clock minutes**.
- **(a) Month-to-date:** 792 ÷ 7.68 days × 30 ≈ **~3,093 min/month** (was ~2,971 at the 04:35Z check, +4%).
- **(b) Last 3 days:** all 193 runs fall in the last 3 days (repo is only ~2.28 days old), so (b) ≈ **~7,920 min/month** (was ~7,130, +11%).

Both projections moved up by roughly the same organic rate as the previous two updates (normal PR/CI cadence, not a spike) — not a material change, so no separate owner notification for this one. Still open, still over the 2,000 Free line on both measures and over the 3,000 Pro line on (b). No action taken by the watchdog; this stays an owner decision (plan tier, and whether to shrink the full-suite cost per #32).

---
_Migrated from #119 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Recomputed independently at 2026-09-16T17:04Z per `docs/WATCHDOG-PROMPT.md` check 6, from the actual Actions run history (not `get_workflow_run_usage`, which still returns 0 on this repo).

The repo's Actions history now starts today: `list_workflow_runs` (all workflows, no filter) returns `total_count: 50`, earliest run created `2026-09-16T10:56:01Z` — there is nothing from before the repo's public migration today, so the 792-minute/193-run figures this issue was tracking no longer exist to observe.

Summing `updated_at − created_at` per run (rounded up to the minute) over all 50 runs: **123 wall-clock minutes total**, all within the last ~6h08m.

- **(a) month-to-date** (Sept 1 00:00 UTC → now, 15.71 days elapsed): 123 ÷ 15.71 × 30 ≈ **235 min/month**.
- **(b) last 3 days** (all 50 runs fall inside this window): 123 ÷ 3 × 30 ≈ **1,230 min/month**.

[...]

PR #160, opened 2026-09-16
guard rail: the review-gate trigger rail also pins the issue_comment subscription (#131)

### LINKED ISSUE #75: guard rail: the review-gate trigger rail passes if the issue_comment subscription is deleted
The rail added in #129 (`guard rails > the review gate wakes on everything its rules depend on`) checks that the `pull_request` trigger list still contains every activity type the rules depend on, and that both markers are still wired into the job's `if:`. It does not check that the **`issue_comment` subscription still exists**.

Deleting this block from `.github/workflows/review-gate.yml`:

```yaml
  issue_comment:
    types: [created]
```

leaves the rail green: `types` then holds only the `pull_request` list, all six required triggers are present, and the `if:` still names both markers. Verified by mutation — the rail passes with the block deleted.

Consequence if it ever happens: no comment can re-stamp the status. A block still lands, because converting to draft is a `pull_request` event; the owner's verdict lands nothing, because it carries no companion event. His approval would sit unread, and a rejection on an unlabelled PR would leave a stale green — the two failures #129 exists to prevent.

Not hypothetical in shape: both regressions this rail has already caught were cost-motivated trigger pruning (#127 narrowed the comment filter for cost; the `labeled` trigger was missing in the first place), and the `issue_comment` subscription is the expensive one.

Fix, one line in `tests/unit/guardrails.test.ts` beside the existing assertions:

```ts
expect({ event: 'issue_comment', subscribed: on.includes('issue_comment') })
  .toEqual({ event: 'issue_comment', subscribed: true });
```

(`on` is already the comment-stripped `on:` section, so prose in the header cannot satisfy it.)

---

For the record, the gate merged in #129 was verified end-to-end on a throwaway PR (#130, closed unmerged, branch deleted) against the code now on `main`: no label -> green; `owner-approval` applied -> red via the `labeled` event; a comment mentioning the approval marker mid-sentence -> job woke and re-stamped red; the marker at the start of a comment -> green.

---
_Migrated from #131 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #75.

The rail added in #129 checks that `review-gate.yml`'s `pull_request` trigger list
still carries every event type the review-gate predicate depends on (`opened`,
`labeled`, `unlabeled`, `converted_to_draft`, `ready_for_review`, `synchronize`)
and that both `REVIEW:`/`OWNER:` markers are wired into the job's `if:`. It never
checked that the **`issue_comment:` subscription** — the trigger that re-stamps
the status when a reviewer posts `REVIEW: CHANGES REQUESTED`/`REVIEW: CLEARED` or
the owner posts `OWNER: APPROVED`/`OWNER: REJECTED` — still exists. Deleting that
block from the workflow left the rail green.

## Fix

One assertion beside the existing ones in `tests/unit/guardrails.test.ts`, reading
the same comment-stripped `on:` section the rail already parses:

```ts
expect({ event: 'issue_comment', subscribed: on.includes('issue_comment') })
  .toEqual({ event: 'issue_comment', subscribed: true });
```

## Proved red

Removed the `issue_comment:` block from `.github/workflows/review-gate.yml`
locally, re-ran the test — failed exactly as expected (`subscribed: false`) —
then restored the file (`git status` clean afterwards, confirmed no diff to the
workflow in this PR).

## Tests run

| check | result |
| --- | --- |
| `npm test` | **976 passed** (30 files) — was 975 |
| `npm run build` | ok |
| e2e `--project=mobile` | **52/52 passed** |

Diff touches only `tests/unit/guardrails.test.ts`, so per #96's filter this PR's
CI will skip e2e — ran it here anyway per the rule that a skipped CI step is not
licence to skip it locally. No guard-rail budget moved in either direction.

## Note on CI

This repository's Actions runners are currently down repo-wide (every job that
needs a runner fails at startup with zero steps — tracked in #155, likely the
same cause as #74's Actions budget projection). This PR's own CI run may come
back red or absent through no fault of this diff; the evidence above is what I
could verify locally.

https://claude.ai/code/session_01UgPzhRH6wSoyN4bXEREsqg

---
_Generated by [Claude Code](https://claude.ai/code/session_01UgPzhRH6wSoyN4bXEREsqg)_

---
_Migrated from #272 in the archived private repo (ugurozsahin/sky-academy-private-archive). History rewritten to remove a third-party reference.__


ugurozsahin:
REVIEW: CHANGES REQUESTED

## What's good
The diff faithfully implements issue #75's own proposed fix, proved red by mutation (block deleted → test fails, restored → clean), and the PR's own note about the repo-wide Actions outage doesn't apply here — CI on this head is green.

## What needs to change
The new assertion only checks that the substring `issue_comment` occurs in `on` (the comment-stripped block covering `name:`/`on:`/`concurrency:`/`permissions:`, not just `on:` itself — worth narrowing that too, but it's not the blocking issue):

```ts
expect({ event: 'issue_comment', subscribed: on.includes('issue_comment') })
  .toEqual({ event: 'issue_comment', subscribed: true });
```

It never checks that `types: [created]` survives underneath that key. The sibling check three lines above, for `pull_request`, does the stronger thing — it extracts the `types: [...]` list and asserts membership:

```ts
for (const t of ['opened', 'labeled', 'unlabeled', 'converted_to_draft', 'ready_for_review', 'synchronize'])
  expect({ trigger: t, subscribed: types.includes(t) }).toEqual({ trigger: t, subscribed: true });
```

So a regression that keeps the `issue_comment:` key but narrows its `types:` to, say, `[edited]` (or drops `types:` to default-all, or to something that excludes `created`) leaves this new rail green while the exact failure mode #75 describes — "no comment can re-stamp the status" on a fresh `REVIEW:`/`OWNER:` comment — comes back. That's a real gap, not a hypothetical: both `pr-test-analyzer` and `silent-failure-hunter` (#99) independently landed on this same finding when run against the diff, each pointing at the same inconsistency with the sibling `pull_request` check right above it.

## What I'm asking for
Extract `issue_comment`'s `types:` the same way the `pull_request` block already does, and assert `created` is in it — mirrors the existing pattern in this same file, one small addition. Proving it red the same way (narrow `types:` locally, watch it fail, restore) closes the loop the PR's own mutation testing didn't cover.

Marking this a draft in the meantime. Ping when pushed and I'll (or whoever picks it up will) re-review — should be quick given how small the diff is.

---

ugurozsahin:
OWNER: APPROVED

ugurozsahin:
Pushed `0435a92`, addressing the blocking review's request: extracted `issue_comment`'s own `types:` list (scoped from the `issue_comment:` key to `concurrency:`, the same way the sibling `pull_request` check is scoped) and asserted `created` is a member of it, rather than only checking that the substring `issue_comment` appears somewhere in `on`.

Proved it red first: locally narrowed `review-gate.yml`'s `issue_comment.types` to `[edited]`, watched the new assertion fail with exactly the shape the reviewer described, restored the real file, confirmed green. `npm test` (976/976), `npx tsc --noEmit` (clean), `npm run build` (ok) — diff touches only `tests/unit/guardrails.test.ts`, none of #96's e2e-trigger paths, so e2e is correctly skipped by CI here too.

This doesn't clear the review — that's for whoever set the original block, or a fresh reviewer, per this repo's protocol. Ready for re-review.

---

---

ugurozsahin:
REVIEW: CLEARED

Clearing this PR under the owner's direct, live authorization for a fresh review — not under the normal #161 adoption rule, which does not apply here: the original blocking comment (2026-09-16T12:43:57Z) carries no `session_<id>` URL, so an ordinary reviewing agent could not legitimately adopt it. The repository owner asked, live in a chat conversation right now, for a new review session to be stood up specifically for PR #160, and this is that review, conducted from scratch rather than as a rubber stamp of the intervening fix.

## Verified the blocking comment's ask against commit 0435a92

The block asked for: extract `issue_comment`'s own `types:` list (scoped the same way the sibling `pull_request` check already is) and assert `created` is a member of it, instead of only checking that the substring `issue_comment` occurs somewhere in `on`. Read the diff, not just the commit message — `0435a92` does exactly that:

```ts
const issueCommentBlock = on.slice(on.indexOf('issue_comment:'), on.indexOf('concurrency:'));
const issueCommentTypes = [...issueCommentBlock.matchAll(/types:\s*\[([^\]]*)\]/g)]
  .flatMap(m => m[1].split(',').map(t => t.trim()));
expect({ event: 'issue_comment', type: 'created', subscribed: issueCommentTypes.includes('created') })
  .toEqual({ event: 'issue_comment', type: 'created', subscribed: true });
```

Reproduced the "prove it red" claim myself rather than trusting it: locally narrowed `review-gate.yml`'s `issue_comment.types` to `[edited]`, re-ran the test — the new assertion failed exactly as described (`subscribed: false`), and, importantly, the *old*, weaker check (`on.includes('issue_comment')`) stayed green under that same mutation — confirming the gap the reviewer flagged was real and is now closed. Restored the file (`git status` clean), full suite green again.

Also checked the `issue_comment:` → `concurrency:` slice against the actual file on this branch: the block order is `pull_request` → `issue_comment` → `concurrency` → `permissions` → `jobs`, so the scoping is correct and doesn't accidentally pull in the `pull_request` types. Deleting the `issue_comment:` key entirely (the original #131/#75 mutation) is still caught — both by the pre-existing substring check and, independently, because the block-slice then evaluates to `''` (fails safe, not silently green).

## Checks run on this head (0435a92)

- `npm test` — 976/976 passed, 30 files
- `npx tsc --noEmit` — clean
- `npm run build` — clean (vite + sw build)
- Diff is `tests/unit/guardrails.test.ts` only — none of #96's e2e-trigger paths (`src/`, `index.html`, `public/`, `tests/e2e/`, `playwright.config.*`, `package*.json`, `ci.yml`), so e2e is correctly skipped by CI on this PR; not run locally either, per that same rule.

## Review agents

No Task/Agent tool was available in this session to actually invoke `pr-test-analyzer`, `silent-failure-hunter` or `type-design-analyzer` as subagents — saying so by name rather than presenting my own reasoning as their output. Applied their published criteria directly instead:
- **Test coverage** (pr-test-analyzer's lens): the new assertion is behavioural (parses the real YAML block, not an implementation detail) and DAMP — it reads like the incident it prevents. No gap rated ≥5 found beyond the scoping question the original block raised, which is now fixed.
- **Silent failures** (silent-failure-hunter's lens): nothing — the diff is a single `expect()` addition to a test file; no try/catch, no fallback, no swallowed errors.
- **Type design** (type-design-analyzer's lens): nothing — no new types introduced.

## Four merge-blocking checks

1. CI green on the current head SHA — yes: run 35195473939 (`test`, `branch-name`) completed/success on `0435a92`, the newest run on this exact head.
2. Latest review requests changes — yes, that was this block; superseded by this comment.
3. Draft / `REVIEW: CHANGES REQUESTED` with no later `REVIEW: CLEARED` — this comment is that clear; the PR is already not draft.
4. `owner-approval` label with no `OWNER: APPROVED` — not labelled `owner-approval` (confirmed independently: the `review-gate` status's block reasons never cited an owner-approval hold, only the open review).

## Decision

The blocking comment's concern is genuinely resolved on the code, not just in the commit message. Merging (squash).

---

Issue #115, opened 2026-09-16
#161's block-adoption condition is unsatisfiable in practice — both open PRs are permanently stalled


ugurozsahin:
**Owner decision needed.** Changing this rule is a **loosening** (it widens when a run may clear another reviewer's block), so no run may merge it. Filing the evidence and a proposal; the label is `owner-input`.

## The observation

#161's escape hatch — "a stale review block may be adopted" — has a second condition that, as written, **can essentially never be satisfied**:

> the session that set it has commented nowhere in the repository since

and the mechanical version in `docs/ROUTINE-PROMPT.md` STEP 2:

> Any comment carrying the same id after the block means that session is still working — the block is live, and not yours to touch, **whatever its age**.

A reviewing run does not stop after posting a block. It posts the block, then goes on to merge another PR, comment on an issue, or open its own. The instant it does, its block becomes permanently unadoptable — not "for now", but for ever, because "since" has no window and the rule explicitly overrides age. The session then ends (scheduled runs are ephemeral) and nobody is left who is allowed to clear it.

So the condition does not measure "is this reviewer still alive". It measures "did this reviewer do anything else at all before finishing", which is ~always true.

## It is not hypothetical — both open PRs are stuck right now (checked 2026-09-12T06:38Z)

| PR | block set | by session | that session commented again at | adoptable? |
| --- | --- | --- | --- | --- |
| #204 (#142) | 2026-09-12T00:57:06Z (~5h41m) | `session_013i4y6JwKq1LfZzN989nEnR` | 02:53:23Z on PR #209 | **no** |
| #209 (#205) | 2026-09-12T02:47:47Z (~3h51m) | `session_012Y8FwsRLFGEx8L8DM3KDEZ` | 02:50:25Z on issue #112, 03:04:01Z on PR #210 | **no** |

Both authors pushed their fixes hours ago (`00970ec` at 01:02Z, `4df2e6d` at 02:53Z). Both PRs are drafts with a red `review-gate`. Neither can move.

#212 is the watchdog's report of the #204 half. This issue is the cause behind it, and #209 fell into the same trap 110 minutes later — which is the point: it is not a one-off, it is the default outcome of every review that blocks.

## Why the rule is shaped this way, and what it was reaching for

[...]

Issue #99, opened 2026-09-16
Agent skills: vendor four superpowers skills, three review agents, a gated frontend-design, and five project skills


ugurozsahin:
Owner (session, 2026-09-10): **add a curated set of agent skills to the repository** — four vendored from outside, three review agents vendored from the official plugin marketplace, five written for this project — so every run (routine, watchdog, owner session) works from the same playbook. Each item below was approved one by one; nothing else goes in.

## Why, and the constraint that shapes how
- A routine run reported that skills named `babysit` and `steward` were missing. No such skills exist in the repo, in Claude Code, or in the official marketplace; the run was reaching for something it expected to have. This issue gives runs the things they actually need instead.
- **Cloud sessions and routines only see three sources of skills**: `.claude/skills/` in the cloned repo, plugins declared in `.claude/settings.json`, and skills enabled on the owner's claude.ai account. The docs add that a plugin from an external source "doesn't load until the team member installs it" — an interactive step no routine can take. So **everything is vendored into the repo**: `SKILL.md` files under `.claude/skills/<name>/`, agent definitions under `.claude/agents/<name>.md`. No `enabledPlugins`, no marketplace, no install step.
- **Context cost is per skill, not per byte.** Only the frontmatter `description` line is in context each turn; the body loads when the skill triggers. Keep descriptions to one line. That is why the count is capped at what is listed here.

## Vendored (copy verbatim, pin the source)
Each vendored file gets a header comment: source repository, path, commit SHA, licence. Do not edit the bodies; if one needs project-specific adjustment, that belongs in a project skill (below) that refers to it.

| Skill | Source | Pin | Licence | Why |
| --- | --- | --- | --- | --- |
| `verification-before-completion` | `obra/superpowers` `skills/verification-before-completion/` | `b36e0829c6d0` | MIT | "Green" is a claim you make after running the command and reading the output — the deleted-e2e-helpers incident (`tsc` and Vitest both blind to `tests/e2e`) would not have shipped |
| `using-git-worktrees` | `obra/superpowers` `skills/using-git-worktrees/` | `b36e0829c6d0` | MIT | Three agents in one checkout corrupted each other's work on 2026-09-10; every parallel agent gets its own worktree |
| `systematic-debugging` | `obra/superpowers` `skills/systematic-debugging/` | `b36e0829c6d0` | MIT | Reproduce → isolate → root cause → fix; matches "prove the rail red before you make it green" |
[...]

PR #161, opened 2026-09-16
Grown-ups screen: a guarded "Start again" that resets the player (#115)

### LINKED ISSUE #71: Grown-ups screen: a guarded "Start again" that resets the player
Owner (2026-09-07): the grown-ups screen should let a grown-up reset the player. **Owner's decision: one single "Start again" that clears everything and returns to onboarding** - no partial-reset variants.

## Why
Half of this already exists as a developer path: `src/main.ts:45` clears `localStorage` when the URL carries `?reset`. No parent will ever find that. And because the game deliberately has **no accounts**, a full reset is the only way to hand the tablet to a second child - so this is less "wipe my data" than "let the next child start". The grown-ups screen, behind the existing gate, is the right home for it.

## What to build
- A clearly separated **"Start again"** action at the bottom of the dashboard (`src/ui/parents.ts:86`), visually distinct from the read-only summary above it and worded for a parent, not a child.
- **Its own confirmation step, not just the gate.** The confirm must require a deliberate act - typing `RESET` (or an equivalent that a child cannot pass by tapping) - and must state plainly what is lost: progress, stars, coins, stickers, streak, name and ninja.
- **Use the existing `.overlay` modal** (`src/style.css:308`), never `window.confirm`/`alert`: a native dialog looks foreign in this game and blocks Playwright, so it would break the e2e run.
- After the reset, go to onboarding - today's avatar screen, or the wizard from #111 once that lands - never to a home screen rendering an empty profile.
- Clear the save through the storage module rather than a raw `localStorage.clear()`, so anything the app keeps outside the save blob is handled in one place and the reset stays correct as storage grows.
- **Offer an undo while the grown-up is still on the screen:** keep the old save in memory and show a short "Undone?" / restore affordance until the screen is left. Cheap to build and it covers a mistaken tap.

## Why the confirmation cannot lean on the gate alone
`gateChallenge()` (`src/game/parents.ts:10`) asks a 6-9 times-table question. That is genuinely out of reach for Reception/Year 1/Year 2 - Year 2 only meets the 2, 5 and 10 times tables - so the gate is sound **today**. But `docs/ROADMAP-PRIMARY.md` extends the game to Year 6, where a child is taught every table to 12x12 by this very game. An irreversible delete must not sit behind a barrier that the product is actively teaching children to pass. Hence a second, non-mathematical confirmation.

## Acceptance criteria
- The reset is reachable only after the gate, requires the explicit typed confirmation, and cannot be triggered by a single tap or a stray Enter key.
- After resetting: the save is back to defaults, the app is on the onboarding screen, and nothing from the previous child (name, avatar, coins, stickers, streak, dojo, personal bests) survives a page reload.
- Cancelling the confirmation changes nothing at all.
- The undo restores the previous save exactly while the grown-up is still on the screen; it is gone once they leave (documented in the UI, not silently).
- No native `alert`/`confirm`/`prompt` anywhere in the flow; the e2e suite runs unattended without dialog handling.
- e2e: gate -> reset -> confirm -> lands on onboarding with an empty profile; and a cancel path that leaves a seeded save untouched.
- Unit test on the reset function itself: it returns the default save shape for every key, and its result passes `migrate()` (guard rail #38).

## Notes
- The `?reset` URL path can stay as a developer convenience, but the UI must not depend on it.
- This is feature work: under the *code health before features* rule it waits behind any open `review` / `debt` issue unless the owner says otherwise.

### Workflow (minimum scrum - the developer and the reviewer/QA must be different agents)
- [x] **Refine** - acceptance criteria written at the top of this issue (Claude proposes, owner may adjust)
- [x] **Develop** - agent A: branch `feature/115-guarded-start-again`, implement + tests (`npm test`, `npx tsc --noEmit`, e2e mobile + desktop), open a PR that says `Closes #115` — PR #274
- [ ] **Review** - a *different* agent reviews the PR diff against CLAUDE.md, the acceptance criteria and the tests; approves or requests changes in the PR
- [ ] **QA** - the reviewer (never the developer) runs the full e2e (mobile + desktop) and walks both the reset and the cancel path on a seeded save (screenshots); results recorded in the PR
- [ ] **Owner action** - approve the wording of the warning and the confirmation
- [ ] **Done** - merged to `main`, issue closed with the commit hash

---
_Migrated from #115 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #71.

The owner's decision on the issue: one single "Start again" that clears everything and returns to onboarding — no partial-reset variants.

## What changed

- A **"Start again"** section at the bottom of the grown-ups dashboard (`src/ui/parents.ts`), below the existing "Move to another device" block and visually separated from it, worded for a parent rather than a child.
- Tapping it opens the existing **`.overlay`** modal (`src/style.css:308`) — never a native `confirm`/`alert`/`prompt` — asking the grown-up to **type `RESET`** to confirm. The "Start again" button in the modal stays disabled until the typed text matches exactly. The modal states plainly what is lost: progress, stars, coins, stickers, streak, name and ninja.
- Cancelling closes the modal and changes nothing — `reset()` is only ever called from the confirm button.
- On confirm: the current save is kept in memory, `reset()` clears it through the storage module (not a raw `localStorage.clear()`), and the modal switches to an **Undo** offer.
- **Undo**, while still on this screen, calls `save(snapshot)` — which replaces every key — restoring the previous save exactly and redrawing the dashboard as if nothing happened.
- Leaving the screen after a reset that was **not** undone — by "Continue" in the modal, or the header's back arrow — always goes to the **avatar/onboarding screen**, never to the map with an empty profile. Only Undo escapes the post-reset state; any other way of leaving commits it.

## Guard rail

`the grown-ups reset flow never uses a native dialog (#71)` in `tests/unit/guardrails.test.ts`: asserts `src/ui/parents.ts` never calls `confirm(`/`alert(`/`prompt(`. Proved red by mutation (temporarily added a `confirm()` call, watched it fail, restored the file — `git status` clean afterwards).

## Found and fixed while screenshotting

The shared `.overlay` class is `position: absolute` sized to `#app`, which works for the play/memory screens (they never scroll past one viewport) but is wrong for the grown-ups dashboard, which is long enough to scroll — the modal centred on the whole document rather than the visible viewport, so it could render off-screen depending on scroll position. Fixed by scoping `position: fixed` to `#reset-overlay` specifically, rather than changing the shared `.overlay` class, so play/memory keep their existing (already-correct) behaviour untouched.

## Tests

| check | result |
| --- | --- |
| `npm test` | **978 passed** (was 976) |
| `npm run build` | ok |
| e2e `--project=mobile` (full suite) | **54/54 passed** |
| e2e `--project=desktop` (grown-ups tests) | **5/5 passed** — ran beyond the mobile-only default because the fixed-position overlay fix is viewport-sensitive |

Two new unit tests in `tests/unit/storage.test.ts` pin `reset()` itself: it returns every key to its default value and the result already passes `migrate()` unchanged, and a `save(snapshot)` taken before a reset restores it exactly (the logic Undo relies on). Two new e2e tests in `tests/e2e/game.spec.ts` cover the typed-confirmation gate + cancel path, and the confirm → Undo → confirm → Continue → onboarding-with-empty-profile path (checked directly against `localStorage`, not via a reload — `seedPlayer`'s init script re-seeds an empty slot on every navigation, which would mask the very thing that assertion checks).

## What was deliberately not done

- **`?reset` is untouched** — the issue allows it to stay as the developer convenience it already is; the UI does not depend on it.
- **No "your progress is not being saved" notice.** That is PR #264's read-only-latch case (a save from a newer build), a different issue with a different trigger; nothing here leaves the app in a silently-not-saving state.
- **Accessibility beyond what the existing `.overlay`/`.gate-input` patterns already provide** — the modal reuses those, so focus goes to the typed-confirmation input on open (matching the existing gate's own `input.focus()` pattern) and every control is a real `<button>`/`<input>`, but no new ARIA beyond what those existing patterns carry.

No guard-rail budget moved in either direction except the one new rail this PR adds (0 → 1, its own).

https://claude.ai/code/session_01ToSrrsLdFSjFkht3vgEk1L

---
_Migrated from #274 in the archived private repo (ugurozsahin/sky-academy-private-archive). History rewritten to remove a third-party reference.__


ugurozsahin:
**Review: approved**, after resolving a merge conflict and independently tracing two findings that turned out not to be reachable. Writing this up in full since the PR arrived as a draft with no `REVIEW:` comment attached (see note at the end).

## Merge conflict
Landing this after #158/#159 conflicted in `tests/unit/storage.test.ts` (adjacent `describe` blocks needing separators, no logic overlap). Resolved, then re-validated on the merged head `5e0f4ed`:

| check | result |
| --- | --- |
| `npx vitest run` | **1001 passed** (30 files) |
| `npm run build` | ok |
| e2e `--project=mobile` (full suite) | **55/55 passed** (3.4m) |
| e2e `--project=desktop` (grown-ups/reset tests) | **6/6 passed** — ran desktop too, the fixed-position overlay fix is viewport-sensitive |

## Read against the diff, CLAUDE.md and #71's acceptance criteria
Typed-word confirmation, `.overlay` reused rather than a native dialog (own guard rail added and proved by mutation), reset routed through the storage module rather than raw `localStorage.clear()`, Undo via `save(snapshot)`, and all three exit paths (back arrow, Continue, hardware/browser back via `popstate`) checked against the "always lands on onboarding, never the map with an empty profile" invariant. This already includes the hardware-back-button fix a prior review round caught (referenced in the PR body as "#274" — an old number from before the repo migration, pointing at this same PR).

**Not `owner-approval`-labelled, and I'm treating that as correct**, not an oversight: CLAUDE.md's owner gate is for a *new look* — art, skins, a redesigned effect. This reuses `.overlay`, `.gate-input`, `.btn` patterns with one new colour variant (`.btn.bad`) for a destructive action, and the copy is plain and functional. #71's own checklist has a stale "Owner action: approve the wording" line from before the 2026-09-10 process — the same kind of superseded checklist line #70 had, and CLAUDE.md is explicit that this class of decision is now the developer's to make and the reviewer's to verify. I read and checked the actual wording (the modal states exactly what is lost, the typed-`RESET` gate, the Undo copy) — it's clear and matches the acceptance criteria; nothing here needs the owner's eyes.

## Vendored agents (#99) — two findings traced and found not reachable
- **pr-test-analyzer**: coverage of the three stated exit paths is solid, e2e checks the actual `localStorage` key post-continue rather than just DOM state. Raised three points, none blocking after checking the code: (1) a tab close/refresh mid-pending-reset — traced `main.ts:61`, `if (load().avatar) nav.map(); else nav.avatar();` on boot already routes correctly regardless of the in-memory `pendingReset` flag, since `reset()` writes through to storage synchronously; (2) fresh-session default of `pendingReset` — module-level `let false`, trivially true on load, not worth a dedicated test; (3) interaction with the `play`/`memory`/`shop`/`parents` `history.back()` branch — see next point, traced as unreachable.
- **silent-failure-hunter**: flagged as "CRITICAL" that `pendingReset` could leak if a back-navigation from `parents` landed on `island`/`rewards`/etc instead of hitting the `isPendingReset()` branch. Traced this: `nav.parents()` is wired from exactly one place (`mapScreen`'s `#grownups` footer button, `src/ui/home.ts:77`), and `nav.map()` never pushes a tagged history entry (`src/main.ts:31`) — so the entry immediately before `parents` in the history stack is always the untagged root, never `island`/`rewards`. The scenario isn't reachable through this app's actual navigation graph. Also flagged a stale-`snapshot`-on-`wireMove`-redraw risk; traced `.overlay`'s CSS (`inset:0`, `z-index:5`, opaque backdrop) and confirmed it blocks interaction with the dash underneath while the reset modal is open, so the redraw path the finding depends on can't fire mid-flow either. Verified rather than taken on trust, per "findings are input to review, never its verdict" — both are good analysis of the code in isolation, neither corresponds to a state a player can reach.
- **type-design-analyzer**: `pendingReset`/`snapshot` are two separately-managed variables rather than one `{kind:'pending', snapshot} | {kind:'none'}` union, so "pending implies a snapshot exists" is convention, not a compiler fact. Reasonable low-cost improvement for a future touch, not blocking — every current exit path does clear the flag, and that's tested.

No guard-rail budget moved except the new rail's own addition (the native-dialog check), matching the PR's claim.

## On the draft state
This PR arrived already in draft, review-gate reporting `Blocked: the PR is a draft`, but with **zero comments on it** — no `REVIEW: CHANGES REQUESTED`, no session id, nothing to adopt under the #161-rule's four conditions (which need a comment to measure age and silence against). Given the repo went public via a migration today and every other PR's comment history came through empty despite their bodies narrating real prior review rounds, this reads as a migration artifact — the draft flag carried over, the comment that explained it didn't. I did a full fresh review rather than treating it as a live, unreadable block (there was nothing to adopt or defer to), and I'm marking it ready for review as part of clearing my own review, not as an override of anyone else's.

Marking ready for review and squash-merging.

---

PR #165, opened 2026-09-16
STEP 3 priority order: priority:P0 outranks priority:P1 (#157)

ugurozsahin:
Closes #157

**Direction: tightening** — this states an ordering the codebase already used
(`scripts/board-sync.mjs`'s `PRIORITIES` array is `['P0', 'P1', 'P2', 'P3']`)
and adds a rail that checks the two agree. Nothing a run may now do widens:
the same issues remain `routine-ok`, the same drop rules in STEP 3 still apply.

## The gap

`docs/ROUTINE-PROMPT.md` STEP 3's ordering rule said, verbatim, "`priority:P1`
before `priority:P2` before `priority:P3`" and had no answer for `priority:P0`
— not one of the three enumerated levels, and not "no `priority:*` label"
either. Two open, unblocked, `routine-ok` issues (#99, #101) carry `priority:P0`
right now, so two runs reading the same repo state could pick differently,
which is exactly the determinism STEP 3 exists to guarantee. #157 raised this
rather than guessing.

`scripts/board-sync.mjs` already resolves the question — its `PRIORITIES`
array is `['P0', 'P1', 'P2', 'P3']`, i.e. the conventional severity-numbering
reading (0 is more urgent than 1) — so this PR states that same order in
`docs/ROUTINE-PROMPT.md` STEP 3 and `BACKLOG.md`'s label list.

## What changed

- `docs/ROUTINE-PROMPT.md` STEP 3, point 4: `priority:P0` now named ahead of
  `priority:P1`, with a pointer to `scripts/board-sync.mjs`'s `PRIORITIES`
  array as the existing precedent.
- `BACKLOG.md`'s label list: `priority:P1/P2/P3` → `priority:P0/P1/P2/P3`,
  with the same one-line pointer.
- New guard rail in `tests/unit/guardrails.test.ts` (`STEP 3 states where
  priority:P0 sorts (#157)`): asserts the STEP 3 sentence names P0 ahead of
  P1, and that `scripts/board-sync.mjs`'s `PRIORITIES` array still agrees.

`CLAUDE.md` needed no change — it only ever said "highest `priority:*` first"
generically, without enumerating P1–P3, so it was already compatible.

## Proved red

Reverted the STEP 3 wording back to the old P1-first sentence locally, ran
the new test — failed exactly as expected (`STEP 3 must say priority:P0
outranks priority:P1, or a P0 issue sorts nowhere`) — then restored the fix
(`git status` clean afterwards).

## Tests run

| check | result |
| --- | --- |
[...]

Issue #97, opened 2026-09-16
A run with nothing to review may take a second item, as a second PR


ugurozsahin:
Owner (session, 2026-09-10): **a run that has nothing to review may take a second item** — as a second, separate pull request.

## Why this shape, and not "two items per run"
The measured bottleneck today is not development. Median pull-request open→merge is **1.2 hours** and 64 PRs merged in 4.1 days; what actually stalls work is **review and conflict** — #150 sat blocked for ten hours, #163 came back twice on stale green ticks, and three agents sharing one checkout corrupted each other's edits. Handing every run two items would double the review queue and the collision surface, which is the opposite of the fix.

Filling **idle** capacity does not: if pull requests are waiting, the run reviews and takes one item as today. Only when nothing is waiting does it take a second. The rule is self-limiting by construction — the moment second items produce a backlog, the condition stops being true and the run goes back to reviewing. That property is the point; do not replace it with a counter or a quota.

## The rule
After STEP 2 (review) and STEP 3 (develop one item), a run may take **one** more item if **all** of these hold:

1. **No pull request is waiting for a review this run could do.** "Could do" excludes: PRs this run opened; PRs already blocked by a `REVIEW: CHANGES REQUESTED` this run cannot clear; and PRs held only by an unanswered `owner-approval`. A PR the run is barred from acting on is not work it is skipping.
2. **Time is left in the run.** STEP 3 already says stop starting new work after ~45 minutes; the second item is subject to the same clock, measured from the start of the run, not from the second item.
3. **The second item's file scope is disjoint from the first.** Check the first PR's changed files and pick something that cannot touch them — a different area label is the cheap heuristic (`curriculum` against `platform`, say), the changed-file list is the real test. Two PRs from one run conflicting with each other would be an entirely self-inflicted version of the problem this repo has hit three times today.
4. **The first item finished.** If it was pushed as WIP with `Part of #<n>`, the run does not start another.

**Two separate pull requests, always** — never one PR closing two issues. They are reviewed independently, they can be merged or blocked independently, and one going bad must not hold the other.

## The cost side, stated
Two PRs mean two CI runs. #176 (running e2e only when the change can affect the game) lands first and cuts most of that; #141 already halved the matrix. Say in the PR what the second item costs in Actions minutes at the current rate, since #119 is open and the spending limit is deliberately closed.

## Acceptance criteria
[...]

Issue #65, opened 2026-09-16
Grown-ups dashboard does not fit a portrait tablet; scrolling works poorly


ugurozsahin:
Owner (2026-09-07, playtest on an Android tablet, portrait): the grown-ups dashboard does not fit the screen - items run past the edge. Scrolling is needed to reach them, and the scrolling itself does not work well.

## Symptom
On an Android tablet in portrait the dashboard content overflows the viewport (at least horizontally), and the page has to be scrolled to read it; the scroll feels wrong / does not get you there comfortably. Not observed on the phone.

## Likely causes (to be confirmed, not assumed)
1. **The 600 px breakpoint treats a portrait tablet as a desktop.** `src/style.css:376` - `@media (min-width: 600px) { .p-stats { grid-template-columns: repeat(4, 1fr); } .p-years { grid-template-columns: repeat(3, 1fr); } }`. A portrait tablet is ~800 px wide, so it gets the four-up stat row and the three-up year grid that were designed for a 1280 px desktop, inside a `.screen` capped at 820 px. Each year card ends up ~250 px wide.
2. **A non-shrinkable child then forces horizontal overflow.** `.p-year-row span { min-width: 130px }` plus the `.isl-bar` and the row gap cannot fit a ~250 px card, so the card's content is wider than the card, and the page gains horizontal scroll. `.p-topic` (icon + flexible text + accuracy) and the `.p-table` header row are the other candidates.
3. **Nested scrollers fight the page scroll.** `.p-table-wrap { overflow-x: auto }` is a horizontal scroller inside a vertically scrolling page; on Android a touch starting inside it is easily captured, which matches "scrolling does not work efficiently". `html, body { overscroll-behavior: none }` (`src/style.css:10`) makes that worse at the boundaries.
4. **`--vh` may be stale.** `src/main.ts:49` sets `--vh` from `window.innerHeight` on `resize` only - not on `orientationchange` or `visualViewport` resize. `#app { min-height: calc(var(--vh) * 100) }`, so a stale value gives the page the wrong height. Worth checking, but it is unlikely to be the main cause here.
5. **No tablet in the test matrix.** `playwright.config.*` runs only `iPhone 13` and `Desktop Chrome 1280x800`, so nothing in CI ever renders a portrait tablet. This is the same coverage gap behind #107.

## Acceptance criteria
[...]

PR #166, opened 2026-09-16
Sound Hunt: stub a working speech engine so the spoken-path test stops racing real time (#131)

### LINKED ISSUE #131: Sound Hunt's spoken-path test races a 4-second real-time voice probe, so it fails intermittently

ugurozsahin:
Closes #131

**Second item this run** (#97) — no open PR was waiting for a review this
run could do (#160 is blocked by a comment with no session URL, so it is
unadoptable per #161 and left for the owner; #165 is this run's own first
item), time was well inside the ~45-minute clock, this PR's only changed
file (`tests/e2e/game.spec.ts`) is disjoint from #165's (`BACKLOG.md`,
`docs/ROUTINE-PROMPT.md`, `tests/unit/guardrails.test.ts`), and #165 had
already been pushed as a finished item (not WIP).

## The bug

`Sound Hunt: nothing to read on the card…` doesn't stub `speechSynthesis`
at all, so it depends on the real browser engine actually starting a
line before `VOICE_START_MS` (a real 4 s wall-clock budget, deliberately
*not* scaled by the suite's speed multiplier — it measures the device, not
a game beat, per the comment in `src/audio.ts`). Headless Chromium in this
container exposes `speechSynthesis` but never actually starts a line
through it, so the probe just counts down; if the test's own actions (goto,
render, answer, pause, quit, re-navigate) take long enough on a loaded
runner, the verdict flips to `no` before the first assertion and the card
shows keywords instead of "🔊 Listen!" — exactly the failure #131 diagnosed
as a timing race, not a real regression.

The issue's own acceptance criteria named three ways to fix it: scale the
budget, inject it under test, or assert the path the run is actually in.
Scaling was explicitly ruled out by the constant's own contract (measures
real device timing, not scaled by `__SNA_FAST`). This PR does the third:
stubs a **working** engine (the mirror of the existing `stubSilentEngine`)
whose `speak()` fires `onstart` synchronously, so the probe verdicts `yes`
immediately and the test no longer depends on real elapsed time at all —
proven by every repeated run below taking well over 4 s end-to-end and
still passing.

This also closes criterion 4 (the second half was vacuous once the verdict
had already flipped to `no` by coincidence): a new assertion pins the
stored voice as `yes` immediately before the `#spk` toggle, so the keywords
shown afterwards are demonstrably caused by turning read-aloud off.

## What changed

- `tests/e2e/game.spec.ts`: added `stubSpeakingEngine` beside the existing
  `stubSilentEngine`/`storedVoice` helpers (moved a few lines earlier so
  the fixed test can use them), called it in the Sound Hunt spoken-path
  test, and added the `storedVoice(page)).toBe('yes')` assertion.
- No production code touched — `src/audio.ts`'s `VOICE_START_MS` and its
  "not scaled" contract are unchanged.

## Proved

Ran the fixed test (and its `#65` no-voice siblings, to confirm nothing
there regressed) 3× each — 15/15 passed, with the fixed test itself taking
4.0–6.4 s per run (i.e. past the old 4 s race window) without failing,
which is the evidence the race is gone rather than just narrowed.

## Tests run

| check | result |
| --- | --- |
| `npm test` | **1003 passed** (unchanged — this PR touches no unit-tested code) |
| `npm run build` | ok |
| `npx playwright test --list --project=mobile` | compiles, still 55 tests (`tests/e2e` is outside `tsconfig`'s `include`, so this is the compile check) |
| e2e `--project=mobile` (full suite) | **55/55 passed** |
| e2e Sound Hunt + `#65` no-voice tests, `--repeat-each=3` | **15/15 passed** |

Diff touches only `tests/e2e/game.spec.ts`, which is inside #96's e2e-reachable
path list, so this PR's CI will run e2e mobile as normal.

No guard-rail budget moved.

https://claude.ai/code/session_01MPYidenUmUPBwwe8gxvcEz

---
_Generated by [Claude Code](https://claude.ai/code/session_01MPYidenUmUPBwwe8gxvcEz)_


ugurozsahin:
REVIEW: CLEARED

Reviewed on a fresh checkout (`git diff main...fix/131-sound-hunt-flaky-voice-probe`).

## What's good
Correctly diagnosed and genuinely eliminates the race rather than narrowing it: `stubSpeakingEngine` fires `onstart` synchronously inside `speak()`, so the probe verdict is settled in the same call stack instead of racing `VOICE_START_MS`'s real 4s wall-clock budget against the rest of the test's actions. The stub still routes through the real production listener (`u.onstart?.()`), so it would catch a regression in the probe wiring itself, not just mask one. The added `storedVoice(page)).toBe('yes')` assertion makes the precondition explicit. The moved-up `stubSilentEngine`/`storedVoice` helpers are unchanged in behaviour (hoisted `const`s, still used correctly by the `#65` no-voice tests below — confirmed those bodies are untouched in the diff), and the timeout/no-voice branch of the probe stays covered separately by those same tests.

## Verified locally
| check | result |
| --- | --- |
| `npm test` | 1001 passed (30 files) — this PR touches no unit-tested code |
| `npm run build` | ok |
| e2e `--project=mobile`, targeted (Sound Hunt + #65 siblings) | 2/2 passed |
| e2e `--project=mobile`, full suite | **55/55 passed** (3.2m) |
| CI (head `5d8433c`) | green — `review-gate`: "No open review block" |

Diff touches only `tests/e2e/game.spec.ts`, inside #96's e2e-reachable path list, so CI correctly ran e2e mobile on this PR.

## Vendored review agents (#99)
- **pr-test-analyzer**: nothing blocking — confirmed the race is genuinely gone (no timer dependency left) and the later `#65` test still binds correctly to the hoisted helpers. One non-blocking nit: `expect.poll` on `storedVoice` could be a plain `expect(await ...)` since the value is settled synchronously now — cosmetic only.
- **silent-failure-hunter**: nothing — the stub goes through the real `onstart` listener rather than bypassing the code path, so it wouldn't mask a real regression in probe wiring; the timeout branch stays covered by the separate silent-engine tests.
- **type-design-analyzer**: nothing — test-helper object literal, no domain types in scope; the one type cast (`onstart as ((e: Event) => void) | null`) is a narrowly-scoped, pre-existing-style test cast, not `as any`.

No blocking findings from any pass. Merged.

Commit: 8dbeaae

---

PR #170, opened 2026-09-16
Add the qa-screenshot project skill (#99)

ugurozsahin:
Part of #99

One slice of #99's remaining "project skills" work: the `qa-screenshot` skill (item 4 in that issue's list).
It documents the bounded visual-QA tooling that already exists in `scripts/` — `shot.mjs` + the `flow-*.mjs`
navigation scripts — but has never been written up anywhere until now.

## What this is

- A catalogue of the existing `flow-*.mjs` scripts and the screen each reaches.
- How to run `shot.mjs` against a built preview server (`npm run build && npm run preview`, matching the port
  `playwright.config.ts` already uses).
- The bounds from #99's own wording: `--project=mobile`'s viewport only, only the screens the diff touches, at
  most three screenshots, a hard time cap, screenshots as PR/review evidence and never repository content.
- Ties the two `CLAUDE.md` verdicts to concrete practice: same screen/viewport/seeded state on `main` and the
  branch settles "the look must not change" for the reviewer to verify directly; a genuinely new look still
  needs the owner's own eyes via `owner-approval`, not a screenshot standing in for `OWNER: APPROVED`.

## What I deliberately did not do

- Did not vendor `frontend-design` or add the `new-ui` label — separate, larger slice of #99.
- Did not touch `docs/ROUTINE-PROMPT.md`'s prompt-slimming ask in #99 — depends on all five project skills
  existing, and `frontend-design` doesn't yet.

## Verified, not just written

Built, ran `npm run preview`, and ran `node scripts/shot.mjs http://localhost:4173/ /tmp/x.png 390 844
scripts/flow-play.mjs` — it reaches the play screen mid-arena exactly as documented, rather than restating the
existing scripts from memory.

## Guard rail

Added `'qa-screenshot': null` to the `SKILLS` allow-list in `tests/unit/guardrails.test.ts` (#180) — the
existing rail that fails the build if a skill directory exists that the list does not name.

**Proved it red first:** ran the allow-list test with the skill directory in place and the list unedited — the
`dirs('.claude/skills')` assertion failed exactly as expected (`qa-screenshot` present on disk, absent from the
list). Added the entry; green.

No new bespoke rail needed beyond the one-line allow-list edit — the existing `it.each` over `SKILLS`/`AGENTS`
already covers the new file's frontmatter. No guard-rail budget moved in either direction.

## Tests run

| check | result |
| --- | --- |
| `npm test` | **1009 passed** (30 files, was 1006) |
| `npx tsc --noEmit` | clean |
[...]

PR #167, opened 2026-09-16
Add the design-language project skill (#99)

ugurozsahin:
Part of #99

**Direction: none (not a governance file)** — this touches `.claude/skills/design-language/` and the SKILLS
allow-list in `tests/unit/guardrails.test.ts`, not `CLAUDE.md`/`BACKLOG.md`/`docs/ROUTINE-PROMPT.md`/the
workflows, so §6 of `open-pr` doesn't apply.

## What this is

One slice of #99's remaining "project skills" work: the `design-language` skill (item 5 in that issue's list).
It collects the visual tokens and UI constraints that already exist in `src/style.css` and the game into one
file, extracted with citations to the real selectors/values rather than restated from memory — the palette
custom properties, the `clamp()` sizing pattern (`.slot`, `.objs .obj`, `.scales .pan`), the three breakpoints
and why each one exists (including the `#109` two-tier tablet fix), touch-target/text-size/read-aloud/British-
English constraints for a Reception-age child, and the `#27` incident behind "reuse a class before inventing
one" (per-index island classes replaced by data-driven CSS variables).

Per `CLAUDE.md`, this is the brief `frontend-design` will be gated against once it is vendored (still open,
along with `qa-screenshot` and the `new-ui` label). On any conflict between the two, this file wins — stated
in §6 of the skill itself.

## What I deliberately did not do

- Did not vendor `frontend-design` or add the `new-ui` label — separate, larger slice of #99.
- Did not write `qa-screenshot` — separate slice.
- Did not touch `docs/ROUTINE-PROMPT.md`'s prompt-slimming ask in #99 — unrelated to this file.

## Guard rail

Added `'design-language': null` to the `SKILLS` allow-list in `tests/unit/guardrails.test.ts` (#180) — the
existing rail that fails the build if a skill directory exists that the list does not name.

**Proved red first:** moved the new skill directory aside and reverted the list edit — the existing
`dirs('.claude/skills')` assertion failed exactly as expected (`design-language` present on disk, absent from
the list). Restored both; green.

No new bespoke rail needed beyond the one-line allow-list edit: the file's frontmatter is checked by the
existing `it.each` over `SKILLS`/`AGENTS` (now 259 tests, +1 for the new entry), and it also picked up two more
dynamic tests for free — the `.claude/` walk at `describe('#161 adoption paraphrase...')` (`docs = walk('.claude/')`)
runs its two `it.each(docs)` checks against the new file automatically (it discusses neither `REVIEW: CHANGES
[...]

PR #171, opened 2026-09-16
importSave() bricking on corrupted progress/mode fields: make every reader tolerant (#95)

### LINKED ISSUE #95: importSave() accepts a versioned blob with wrong-typed fields, which bricks the app
Found reviewing PR #172 (merged as `96f393a`). Recorded rather than blocked on, because it needs a hand-crafted blob and `load()` has had the same exposure since it was written — but it stops being theoretical once #20's profiles and #16's second player move saves through this code, which `importSave`'s own doc comment anticipates.

## What happens

`importSave()` guards on the version only — `typeof v === 'number'`, integer, `1 <= v <= SAVE_VERSION` — and then hands the blob straight to `migrate()`, which is `{ ...DEFAULT, ...s, v: SAVE_VERSION }`. Anything carrying a plausible `v` is accepted whatever the rest of it looks like. Two cases, both reproduced against `main`:

- **`{"v":1}` is accepted and wipes the save.** `migrate()` fills every missing key from `DEFAULT`, so the result is a fresh player, and the UI reports "Restored — the progress below came from that code." This is exactly the case the version check was written to stop: PR #172's body says "`{}` and `{"coins":999}` are valid JSON that `migrate()` would happily turn into a fresh default", and adding a `v` walks straight past the fix.
- **`{"v":1,"progress":null}` is accepted, persisted to `localStorage`, and then throws.** `parentSummary()` dies on it (`TypeError: Cannot read properties of null`), and so does the map screen — `src/ui/home.ts:53` does `d.progress[t.id]?.stars`, where `?.` guards the *value*, not the null container. Because the blob was already written to storage, every subsequent launch reloads it and throws again: **no in-app way back**, only clearing site data.

## What is not broken (checked, so nobody re-checks it)

The version gate does its job on every realistic stray paste: prose, a URL, a truncated code, smart quotes, `null`, `42`, `[1,2,3]`, `[{"v":1}]`, `{"save":{"v":1}}` are all refused with the existing save untouched; `{"v":0}`, `{"v":1.5}` and `{"v": SAVE_VERSION+1}` are refused; whitespace- and newline-wrapped codes are correctly *accepted*, which is the email-paste case; and `JSON.parse` + spread means no `__proto__` pollution.

## Suggested fix

Validate the shape, not just the version — a small `isSaveShaped(raw)` beside `migrate()` checking that the keys `migrate()` will not fill are the right *type*: the five record fields (`progress`, `endless`, `sprint`, `boss`, `memory`, `training`, `equipped`) are non-null objects, `stickers`/`owned` are arrays, `coins`/`spent` are numbers, `streak` is an object. `importSave()` refuses when it fails.

Two decisions worth making deliberately:

1. **Whether `load()` should use it too.** It has the same hole from a hand-edited or half-written `localStorage`, and the answer there is probably *fall back to `DEFAULT`* rather than refuse — a corrupt device save should still boot the game, where a bad paste should be rejected. Same predicate, different failure behaviour.
2. **Whether a bare `{"v":1}` should be refused outright.** It is shape-valid but semantically "a fresh player", so a caller cannot tell a restore from a wipe. Refusing any blob with no `name` and no `progress` would cover it.

## Tests to add with the fix

- Unit: each wrong-typed field refused, with the existing save asserted unchanged (the pattern `storage.test.ts` already uses).
- **Guard rail:** the one that would have caught this — assert that a blob accepted by `importSave()` can be fed to `parentSummary()` and to `home.ts`'s star tally without throwing. That turns "does it parse" into "is it usable", which is the property that actually matters here.

---
_Migrated from #174 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #95

Taking this as a second item this run (STEP 2 review already done — PR #167 merged, #160 left blocked for the
owner per #161's adoption test; files here are disjoint from this run's first item, PR #170's
`.claude/skills/qa-screenshot/SKILL.md` + `tests/unit/guardrails.test.ts`).

## The bug, reproduced against current `main` first

`importSave()` only checks `v`, so a hand-edited or corrupted "Restore" paste like `{ v: 1, progress: null }`
passes the version check and is written straight through — `migrate()` is `{ ...DEFAULT, ...s, v: SAVE_VERSION
}`, which keeps whatever type `s` carries. Both cases from the issue still reproduce despite #232's later
hardening (`isMigratable()`/`isFutureSave()` only guard the *version*, not the shape):

- a bare `{ v: 1 }` still reads indistinguishably from a wipe;
- `{ v: 1, progress: null }` is still accepted and persisted, and `parentSummary()` and the map screen's star
  tally both throw on the very next launch — `d.progress[id]` cannot be guarded by `?.` when `d.progress`
  itself, not the value at `id`, is `null`.

## Why this fix, not the issue's suggested `isSaveShaped()` gate

The issue's own suggestion was a shape validator that makes `importSave()` refuse a wrong-typed blob outright.
I started there and reverted it: **`tests/unit/storage.test.ts` already has a deliberate, existing test (#270)
that feeds `importSave()` exactly this shape of corruption — `{ progress: null }`, `{ boss: null }`, `{ memory:
'not an object' }`, `{ sprint: [] }` — and asserts it is *accepted*, relying on `evaluateStickers()`/`addCoins()`
reading those fields tolerantly afterwards.** #270's own reasoning (quoted in `storage.ts`) chose "accept the
import, make every reader tolerant" over "reject at the door" for that path. Rejecting at import now would
directly contradict that established, tested decision — and re-fighting it wasn't this issue's ask.

So this is the same fix reaching the readers #270 did not touch, not a new philosophy:

- Exported `storage.ts`'s existing `safeRecord()` helper (previously module-private, already used internally by
  `evaluateStickers`/`wallet()` for exactly this class of blob).
- Used it at every place a `SaveData` record field was indexed directly instead of through a loader function:
  - `home.ts`'s `mapScreen` (`progress`) and `islandScreen` (`progress`, `training`, `endless`, `sprint`,
    `boss`, `memory`);
  - `avatar.ts`'s `avatarScreen` (`progress`, via `masterProgress()`);
  - `game/parents.ts`'s `parentSummary()` (`progress`, `endless`, `sprint`, `boss`, `memory`, `training`) —
    the exact function, and the map screen, the issue names as throwing.

## The issue's two open decisions, made deliberately

1. **Whether `load()` should use the same validation.** Not touched here. #232 already refuses an unreadable
   `v` on load and resets (a shape no build ever wrote has nothing on the other side to preserve); a
   wrong-typed-but-*readable* field reaching an ordinary boot is a narrower, lower-probability case than a
   deliberate "Restore" paste, and conflating the two risks widening this PR into #232's territory. Left as a
   separate decision if the owner wants it.
2. **Whether a bare `{ v: 1 }` should be refused outright.** Not done. A legitimately exported fresh-player save
   has exactly that shape — `DEFAULT.name` is `''`, `DEFAULT.progress` is `{}` — so refusing "no name and no
   progress" would reject a genuine restore alongside the ambiguous one it targets. Left alone.

- Did not touch `load()` (decision 1 above).
- Did not add an `isSaveShaped()` gate to `importSave()` (contradicts #270's existing test, see above).
- Did not widen scope to every `SaveData` field — `coins`/`spent`/`stickers`/`owned`/`streak`/`equipped` are
  read through `wallet()`/other already-tolerant paths or are not indexed the way `progress` and the four
  per-year record fields are; nothing in this diff touches them.

## Tests run

| check | result |
| --- | --- |
| `npm test` | **1010 passed** (30 files, was 1009) |
| `npm run build` | ok |
| e2e `--project=mobile` (full suite) | **56/56 passed** (was 55) |

New tests:
- **Unit** (`tests/unit/parents.test.ts`): `parentSummary()` fed every affected field corrupted
  (`null`/wrong-type/array/`undefined`) in one call — asserts it does not throw and every field reads as empty
  rather than exercising one field at a time.
- **e2e guard rail** (`tests/e2e/game.spec.ts`, `a corrupted save does not brick the app (#95)`): seeds the same
  corrupted blob via `seedPlayer()`'s `extra` param, then walks the map → an island → back → the grown-ups
  dashboard with a `pageerror` listener throughout (not a console-error listener too — that also caught an
  unrelated blocked-font-fetch console error in this sandboxed environment, which is not what this test is
  about). `home.ts`/`avatar.ts` render through `document` and this repo's unit suite has no jsdom, so this is
  the only way to exercise the actual crash sites the issue names, not just the pure function underneath them.

This diff touches `src/`, `tests/e2e/game.spec.ts` and no `package*.json`/`playwright.config.*`/`ci.yml` — so
per #96 this PR's CI **will** run the e2e step (unlike PR #170). Ran it here first anyway, full mobile suite.

https://claude.ai/code/session_01K1g3BvhoDRf5WArAD1TPc3

---
_Generated by [Claude Code](https://claude.ai/code/session_01K1g3BvhoDRf5WArAD1TPc3)_


ugurozsahin:
REVIEW: CHANGES REQUESTED

Session: https://claude.ai/code/session_014Zmk2ZfxgaWVtGeCwHxysY

The tolerant-readers approach over the shape-validator the issue suggested is a reasonable, well-justified call (staying consistent with #270's precedent, explained clearly in the PR body) — that part I have no objection to. `npm test` (1010), `tsc --noEmit`, `npm run build` and `npx playwright test --project=mobile` (56/56) all reproduce clean on the branch. The three vendored agents ran on the diff and found real gaps I then verified directly against the code, not just took on their word:

**1. `d.streak.days` is unguarded and reachable from the very screen this PR claims to fix.** `src/ui/home.ts:23`, inside `topbar()`: `` d.streak.days > 1 ? ... `` with no `?.`/`safeRecord` — `topbar()` is called by `mapScreen()` (and `islandScreen()`), so a corrupted `{ v: 1, streak: null }` Restore paste still throws on the map, the exact failure class #95 is about, in a field this PR didn't touch. Worse: the PR body's own "What I deliberately did not do" section states `streak` is "read through `wallet()`/other already-tolerant paths" — that's not correct for this call site; `wallet()` doesn't touch `streak`, and `topbar()` reads `d.streak` directly. `home.ts:176` (the rewards screen) has the same unguarded `d.streak.days`.

**2. `dojoCard()`, also rendered on the map, is unguarded the same way.** `dojoToday()` (`storage.ts:317`) → `dojoFor(load().dojo, ...)` (`game/dojo.ts:53`) does `s.date === date` on `s` with no null check; a corrupted `dojo: null` throws there too, before `dojoCard()` can render — again on the map screen.

**3. The five `record*()` write functions in `storage.ts` (172, 179, 184, 188, 192, 197, 202 — `recordTopic`, `recordAccuracy`, `recordTraining`, `recordEndless`, `recordSprint`, `recordBossWin`, `recordMemory`) still index `load().progress[id]` / `load().training[year]` etc. directly, unguarded.** This PR's own fix is what lets a player reach the topic list and tap a topic on a corrupted save (previously they'd have crashed at the map first); finishing that topic now crashes here instead. Not mentioned anywhere in the PR's scope notes as a deliberate exclusion.

None of this is exercised by the new tests — the e2e guard rail seeds `progress/endless/sprint/boss/training` as corrupted but leaves `streak` and `dojo` untouched, and never plays a topic to completion, so it passes despite items 1–3 above being live bugs on the same save.

What would resolve this: extend the `safeRecord`-tolerant treatment to `d.streak` (home.ts:23, home.ts:176) and `d.dojo` (storage.ts:317 / game/dojo.ts:53), and to the `record*()` write path in storage.ts (either guard each call site the way the readers were, or — per the type-design-analyzer's suggestion, worth considering since it'd remove the need for every future call site to know about this hazard — coerce the six record fields once inside `migrate()`, which already exists to bring a raw blob up to `SaveData` shape). Add or extend the e2e/unit coverage to actually exercise `streak`/`dojo` corruption and a full record-a-topic round trip on a corrupted save, matching the issue's own suggested guard rail ("a blob accepted by `importSave()` can be fed to \[every reader] without throwing").

---

ugurozsahin:
Pushed `7a6fee2`. Addressed all three findings by centralising the fix rather than guarding more call sites individually — the type-design-analyzer's suggestion was right, and once I looked, findings 1–3 all have the same shape: a wrong-typed field reaching a reader that trusted `load()` to hand it something shaped like `SaveData`.

`migrate()` now runs a `sanitizeTypes()` pass before the `{ ...DEFAULT, ...s }` merge: any of `progress`/`endless`/`sprint`/`boss`/`memory`/`training`/`equipped`/`streak`/`dojo` that isn't a plain object, `stickers`/`owned` that isn't an array, or `coins`/`spent` that isn't a number, gets deleted from the raw blob so `DEFAULT`'s value fills it in — exactly like a missing key already does. Since every reader (`topbar()`, `dojoCard()`, every `record*()` writer, `home.ts`, `avatar.ts`, `game/parents.ts`) goes through `load()` → `migrate()`, this closes findings 1, 2 and 3 in one place, with nothing downstream needing to know the hazard exists. A validly-shaped value is left exactly as given (pinned in the new `migrate()` unit test).

I left the existing `safeRecord()` guards in `home.ts`/`avatar.ts`/`game/parents.ts` from the first cut in place rather than stripping them back out — they're redundant now for anything sourced via `load()`, but `parentSummary()` specifically is called directly with hand-built `SaveData` objects in both the unit test and (deliberately) nowhere else, so its own tolerance is still load-bearing there, not dead code.

Tests:
- **Unit** (`tests/unit/storage.test.ts`, new describe block): `migrate()` drops a wrong-typed `streak`/`dojo`/`equipped`/`owned`/`coins` and keeps a validly-shaped one unchanged; every `record*()` writer + `touchStreak()` + `recordDojo()`/`dojoToday()` survive an all-fields-corrupted `importSave()` without throwing, and the writes actually land (`load().progress['y1-add']`, `load().endless.year1`) — not just "didn't throw."
- **e2e**: extended the existing seed with `streak: null, dojo: null`; added a second test that plays a **full mission to completion** on the corrupted save and asserts `stars`/`coins` actually saved — the round-trip finding 3 named, which a unit test calling the writers directly can't stand in for on its own. (Caught my own bug writing this: the new `describe` block sits outside the suite's main one and doesn't inherit its `window.__SNA_FAST = 4` — the mission test ran real-time first, 56s vs 17s for the identical mission once fixed. Fixed before pushing, not left for you to catch.)

| check | result |
| --- | --- |
| `npm test` | **1009 passed** (30 files) |
| `npm run build` | ok |
| e2e `--project=mobile` (full suite) | **57/57 passed** |

Left as draft — not undrafting my own review block. Over to you.

---

ugurozsahin:
REVIEW: CLEARED

Clearing another reviewer's block, adopted under #161. All four conditions hold: the block (comment at 2026-09-16T20:45:06Z, session `session_014Zmk2ZfxgaWVtGeCwHxysY`) is ~4h11m old at time of writing, over the 4-hour minimum; that session has posted no further comment on *this* pull request since the block itself — its only other trace here would be the fix-push comment at 21:00:08Z, but that comment carries no session URL in its footer, so it is a different session (the author, not the reviewer) and does not protect the block; and I re-derived the original objection against the current head (`7a6fee2`) myself, not from the PR body's own account of it.

**The original objection is genuinely resolved.** It named three unguarded call sites reachable from the corrupted-save class this PR fixes: `d.streak.days` in `topbar()` (`home.ts:23`, `home.ts:176`), `dojoCard()` → `dojoFor()` (`storage.ts:317` / `game/dojo.ts:53`), and the five `record*()` writers in `storage.ts`. The fix centralises this in a new `sanitizeTypes()` inside `migrate()` rather than guarding each site — I mutated it out (removed the `sanitizeTypes()` call, restoring the original bug) and confirmed 4 tests fail with exactly the three failure classes named in the block (`recordTopic` throwing on null `progress`, `streak.days` on null `streak`, etc.); restored and confirmed the suite is green again (1009/1009, `tsc` clean, `build` ok, e2e mobile 57/57 including the two new corrupted-save guard rails). So the objection this block recorded no longer applies to the current head.

That said — see the next comment. This PR is not clear to merge; a related but different gap survived the fix.

---

ugurozsahin:
REVIEW: CHANGES REQUESTED

Session: https://claude.ai/code/session_01MjdALbJJ3auTPdEWqQPKN3

Ran the three vendored review agents on the diff (`pr-test-analyzer`, `silent-failure-hunter`, `type-design-analyzer`). Their findings are input, not the verdict, but `pr-test-analyzer`'s is one I verified myself and it blocks.

**Blocking — `pr-test-analyzer`'s finding, verified: `sanitizeTypes()` only sanitizes object/array/number fields, never the primitive-typed ones, and `avatarScreen()` reads one unconditionally.** `src/ui/avatar.ts:24-25` does `esc(d.name)` and `hasName(d.name)`; `esc()` (`dom.ts:3`) is `s.replace(...)`, and `hasName()` (`avatar.ts:8`) is `name.trim()`. Neither is guarded, and `sanitizeTypes()` (`storage.ts`) never touches `name`, `avatar`, `year`, `voice`, `sound`, `speech` or `tutorialSeen` — it only has branches for the object-shaped, array-shaped and number-shaped fields. I reproduced it directly:

```
migrate({ v: 1, name: 123 }).name  // → 123 (number), not sanitised back to a string
```

A hand-edited or corrupted "Restore" paste with a wrong-typed `name` — exactly the class of input this issue and this PR are about, and a field whose entire point is being freely typed by a human into the Restore box — passes `importSave()`, gets persisted, and crashes the instant the player opens "Change ninja" from the map. That is a screen every returning player can reach, and it is the same failure mode (a reader indexing an untyped field without a guard) this PR exists to close everywhere else. None of the new tests corrupt `name` (both the unit and e2e seeds use `'Ada'`), so nothing here would have caught it.

What would resolve this: extend `sanitizeTypes()` with a string check for `name`/`avatar`/`year`/`voice` and a boolean check for `sound`/`speech`/`tutorialSeen` (mirroring the existing object/array/number branches), plus a unit test pinning `migrate({ v: 1, name: 123 }).name === ''` (or whatever the chosen fallback is) and an e2e/unit case that opens the avatar screen on such a save.

**Not blocking, worth doing in this PR or a fast follow-up:**
- `silent-failure-hunter`: `importSave()` returns `true`/`false` purely off the version check, so a Restore that silently drops several fields via `sanitizeTypes()` still shows the unqualified "Restored" message in `parents.ts` — the one place in this whole flow where telling the adult something was dropped would actually be actionable. Pre-existing pattern from #270, but this PR widens what it silently swallows (`streak`, `dojo`, `coins`, `equipped`, `owned`, `spent`). Suggested: have `importSave()` report whether `sanitizeTypes` changed anything and surface a distinct message.
- `type-design-analyzer`: the three `as const` field-name arrays in `sanitizeTypes()` have no exhaustiveness check against `SaveData` — a new field added to the interface later and forgotten here reintroduces this exact bug with nothing red. Worth a guard-rail test asserting every `Record`/array-typed `SaveData` key appears in one of the lists (which would also have caught this review's finding, since `name` etc. currently appear in none of them — worth deciding whether primitives get a fourth list or their own explicit allow-list too).

Everything else in this PR is solid — see the cleared-block comment above for what I verified. Do not fix this in the same run; leave it for whoever picks it back up.

---

ugurozsahin:
Pushed `7691e14`. Confirmed the finding first — `sanitizeTypes()`'s three branches only covered object/array/number-shaped fields, so `migrate({ v: 1, name: 123 }).name` really did come back `123`, and `avatarScreen()`'s `esc(d.name)`/`hasName(d.name)` really do throw on that.

Extended `sanitizeTypes()` with the primitive fields, same pattern as the existing branches:
- `name`/`year` — must be a string, else dropped.
- `avatar` — string **or** `null` (null is the legitimate "no avatar chosen" value, so it isn't just a `typeof === 'string'` check).
- `voice` — must be one of `'unknown'`/`'yes'`/`'no'`, else dropped.
- `sound`/`speech`/`tutorialSeen` — must be a boolean, else dropped.

Left the two non-blocking suggestions (the "some progress could not be restored" message, the exhaustiveness guard rail over `sanitizeTypes()`'s field lists) untouched — you said explicitly to leave both for later, not fix them in this round.

Tests:
- **Unit**: `migrate()` fed every primitive field corrupted in one call (including `avatar: 42`, which should drop, vs. the earlier test's `avatar: null`, which should not) — asserts each falls back to `DEFAULT`, and a validly-shaped set including `avatar: null` is kept unchanged; a corrupted `name` imported via `importSave()` comes back a string that `esc()`/`.trim()` — the exact `avatarScreen()` calls — don't throw on.
- **e2e**: new guard rail opens the avatar screen via `#change-av` on a save with `name: 123` (plus `sound`/`tutorialSeen` also wrong-typed) and asserts it renders with the name input empty, not literally `"123"`.

| check | result |
| --- | --- |
| `npm test` | **1011 passed** (30 files) |
| `npm run build` | ok |
| e2e `--project=mobile` (full suite) | **58/58 passed** |

Left as draft, per convention — over to you again.

---

ugurozsahin:
REVIEW: CLEARED

Clearing another reviewer's block, adopted under #161. All four conditions hold, checked at 2026-09-17T06:4x:xxZ:
- **Age**: the current block (comment at 2026-09-17T00:56:49Z, session `session_01MjdALbJJ3auTPdEWqQPKN3`) is ~5h44m old — over the 4h floor.
- **Setter silence**: that session has posted no comment on this PR since — the only later comment (01:05:17Z) is the author's own fix-push note, not from the blocking session. ~5h35m of silence on this PR, over the 2h floor.
- **Re-derivation**: the block's objection was that `sanitizeTypes()`'s object/array/number branches never touched `name`/`avatar`/`year`/`voice`/`sound`/`speech`/`tutorialSeen`, so `migrate({ v: 1, name: 123 }).name` came back `123` unchanged and crashed `avatarScreen()`'s `esc(d.name)`/`hasName(d.name)`. Commit `7691e14` adds exactly those branches. I mutation-tested it: reverted the new lines in `sanitizeTypes()`, and the two new unit tests in `storage.test.ts` failed with precisely the described failure modes (`m.name` stayed `123`; `typeof d.name` stayed `'number'`); restored and confirmed green again. The objection is genuinely resolved.
- This comment names the three conditions above, per #161.

Session: https://claude.ai/code/session_019C4e1q2Be1z9wRgSKkiwur

## Full review this run

| check | result |
| --- | --- |
| `npm test` | 1011/1011 passed |
| `npm run build` | ok |
| e2e `--project=mobile` (full suite) | 58/58 passed |

Ran the three vendored review agents fresh on the full diff (`main...7691e14`):
- **pr-test-analyzer**: coverage adequate for the bug this PR fixes. Two minor completeness notes, not blocking (`stickers`/`spent` not directly pinned with a wrong-type case the way `owned`/`coins` are; no test for a right-typed-but-invalid value like `avatar: "nonexistent-id"` — confirmed not a real gap, both are pre-existing tolerant call sites). Confirmed `certs` isn't in `sanitizeTypes()` but `certificates()` already self-guards, so no crash path.
- **silent-failure-hunter**: the silent-drop-and-default pattern is the right call for a kids' Restore flow. Flagged that `save()` will overwrite the original corrupted blob with the sanitized+defaulted one on the next write, with no distinct signal to the grown-up that something was dropped — not new (this is the round-2-deferred "some progress could not be restored" message, explicitly left for a fast follow-up per this PR's own body), but worth its own tracked issue.
- **type-design-analyzer**: flagged that `sanitizeTypes()`'s field lists have no compile- or test-time exhaustiveness check against `SaveData`, and pointed at `certs` as a live instance.

I verified the `certs` point myself: `certs` genuinely isn't covered by `sanitizeTypes()`, but its one read site (`certificates()`) already guards with `Array.isArray`, so nothing crashes today — the gap is real but not blocking. Filed #177 for the exhaustiveness rail, which covers both the type-design-analyzer point and the silent-failure-hunter's deferred-message point can ride alongside it as a separate concern if the owner wants it split out.

None of the three found anything blocking this PR. Merging.

---

---

Issue #101, opened 2026-09-16
Context by layer: CLAUDE.md as a map, .claude/rules/ scoped by path, three PreToolUse hooks, prompts carry flow only


ugurozsahin:
Owner (session, 2026-09-10): **give runs context by layer — what a run loads should depend on what it is touching, not be the same 45 KB for everyone.** This follows Claude Code's documented model (CLAUDE.md → `.claude/rules/` with `paths:` → skills → hooks) and the owner asked for it on the condition that it is the official best practice, which it is: the memory docs say to keep CLAUDE.md under 200 lines of "facts Claude should hold in every session" and to move anything path-specific into rules and anything procedural into skills.

## Where we started (measured 2026-09-10)
- Every run loaded the same set at STEP 1: `CLAUDE.md` **9.8 KB**, `docs/ROUTINE-PROMPT.md` **26.9 KB**, `BACKLOG.md`, `docs/CURRICULUM.md` ≈ **45 KB ≈ 11k tokens** before the first useful action. A one-line docs fix and a new curriculum topic started identically.
- There was **no `.claude/rules/`**, no nested `CLAUDE.md`, no `.claude/settings.json` — no layer between "always" and "the run reads a file itself".

## The layers, and what landed in each

| Layer | Loads when | Mechanism | Ours |
| --- | --- | --- | --- |
| 1 Always | session start | `CLAUDE.md` | Trimmed **18,999 → 10,750 bytes** (PR #198, #203, #222, #235) |
| 2 By path | Claude reads a matching file | `.claude/rules/<topic>.md` with `paths:` frontmatter | **All 7 files exist**: `curriculum.md`, `game.md`, `e2e.md`, `guardrails.md`, `governance.md`, `android.md`, `style.md` — each with a non-empty `paths:` list matching real files, held by a guard rail |
| 3 By task | skill triggers | `.claude/skills/` | `open-pr`, `review-pr`, `add-guard-rail`, `qa-screenshot`, `design-language`, `add-topic` |
| 4 By role | run start | `docs/ROUTINE-PROMPT.md`, `docs/WATCHDOG-PROMPT.md` | Trimmed **44,034 → 40,949 bytes**; flow only, every "how" line points at layer 2 or 3 |
| 5 By item | issue taken | the issue body | acceptance criteria per item |
[...]

Issue #62, opened 2026-09-16
routine: heartbeat


ugurozsahin:
2026-09-18T04:56Z — reviewed PR #221 (#101): re-derived the original blocking objection (line-anchored `OWNER: APPROVED` check) against the current head and confirmed it fixed, but found a new, separate blocking issue in the same diff before clearing — the strict/loose marker checks are POSIX-ASCII-only, so a body opening with an invisible Unicode character (ZWSP/BOM/NBSP/soft-hyphen, all reproduced against the real hook) bypasses the whole guard; posted a fresh `REVIEW: CHANGES REQUESTED` rather than adopting/clearing the stale one. Developed #77 as PR #226 (`Closes #77`, review-gate edited/deleted triggers). Took a second item: #116 item 1 as PR #227 (`Part of #116`, the build-scripts silent-no-op guard).

- nightly: ok — run 35200970335, 2026-09-17T08:40:45Z, success, head `020948e` (~20h15m old, inside 26h)
- watchdog pulse: ok — issue #61 body timestamped 2026-09-17T23:08:07Z (~5h48m old, inside ~14h)
- board: stale — issue #164 pulse still reads 2026-09-17T15:37:47Z (~13h18m old, past ~2h) — already tracked as issue #220 (P1, watchdog), no new information
- main: green — tip `cc86cbb` unchanged this run (no merges)
- open PRs: 3 — **#221** (not mine; blocked by my own new `REVIEW: CHANGES REQUESTED`, a Unicode-bypass finding unrelated to the now-fixed original objection — not adoptable by anyone until fixed and reviewed fresh), **#226** (mine, `Closes #77`, CI green, draft — ready for a different session's review), **#227** (mine, `Part of #116`, CI green/queued at time of writing, draft — ready for a different session's review)
- second item: yes — #116 item 1 (PR #227). Condition 1 held once #226 was opened (only #221, barred as my own fresh block, and #226, barred as mine, remained open); condition 2 held (~18 min of the ~45-min clock used); condition 3 held (scripts/*.mjs, disjoint from #226's `.github/workflows/review-gate.yml`); condition 4 held (#226 pushed green, not WIP)
[...]

Issue #88, opened 2026-09-16
review: undrafting a PR does not re-run CI, so a blocked PR merges on a stale tick


ugurozsahin:
Found from #150's own lifecycle this morning. Nothing is broken on `main`; this is a hole in the merge gate that stayed invisible because the one time it fired, the tree happened to be fine.

## The defect

`.github/workflows/ci.yml` declares:

```yaml
on:
  pull_request:
    paths-ignore: [...]
```

with **no `types:` list**, so GitHub applies the default `[opened, synchronize, reopened]`. **`ready_for_review` is not among them.**

The review protocol in CLAUDE.md blocks a PR by *converting it to a draft*. Clearing the block marks it *ready for review*. That transition fires no CI. So the sequence the protocol itself prescribes — block → draft → author fixes → reviewer clears → merge — ends with a merge decided on whatever CI run last happened, however old.

## What it did to #150

| | |
|---|---|
| blocked, converted to draft | 00:46Z |
| author's fix pushed (`3c3922d`), CI green | 00:57Z |
| last CI run on the head `d8e8aa0` | **03:31Z** |
| `main` advanced: #152, #153, #154, #155 | 03:31Z → 12:32Z |
| undrafted + `REVIEW: CLEARED` | 12:34Z — **no CI run fired** |
| merged | 12:37Z, on the 03:31Z tick |

Nine hours and four merges of `main` separated the evidence from the decision. **#155 rewrote the e2e helpers (`pickAvatar`/`startTopic`) that #150's new guard rail calls**, so that tick was not evidence about the tree being merged — the reviewing run said as much, said it would wait for a re-run, and merged when the re-run it was expecting never arrived. It could not have arrived.

It came out fine: I merged `main` into the branch independently and validated the merged tree — tsc clean, unit **503/503**, build 152.93 kB, the #30 rail green on mobile and desktop against #155's rewritten helpers, and still failing correctly when the rule is removed. But that was a separate manual check, not the gate doing its job.

## Why this is worth fixing rather than remembering

"Never merge a PR whose CI is not green" is the first merge rule, and it is read as *green*, not *green on this tree*. A draft that sits for hours is the normal case here, not an edge case — that is what a review block **is**. So the gap is on the protocol's happy path.

[...]

PR #178, opened 2026-09-17
Layer 2: a .claude/rules/curriculum.md rule file, scoped by paths (#101)

ugurozsahin:
Part of #101

**Tightening.** Adds a guard rail (a `.claude/rules/*.md` must carry a valid `paths:` list) and removes no
constraint; loads nothing new unconditionally (a Layer 2 file only loads when a session touches a matching
path, the opposite of adding to what every turn already pays for CLAUDE.md).

## What changed and why

First slice of #101 (context by layer). A previous run's "starting" comment on the issue (2026-09-17T02:46:54Z)
proposed this exact slice and then appears to have died before pushing a branch (`git ls-remote` found nothing
matching `101`, and `routine: heartbeat` #62 was not updated after that run's predecessor at 00:58:57Z) — this
PR is that slice, done.

- **New `.claude/rules/curriculum.md`**, with `paths: [src/curriculum/**, tests/unit/curriculum.test.ts]`
  frontmatter, carrying the two curriculum-specific bullets moved out of `CLAUDE.md`'s always-loaded `## Rules`
  (generator/test contract, difficulty-stage/NC-range rules).
- **`CLAUDE.md`** loses those two bullets, replaced with one pointer line. The `window.__sna` bullet stays —
  it documents `src/ui/play.ts`'s e2e contract, not anything curriculum-scoped.
- **Guard rail** (`tests/unit/guardrails.test.ts`, new describe block): every `.claude/rules/*.md` must open
  with YAML frontmatter carrying a non-empty `paths:` list, and every entry in it must match at least one real
  file in the repo (checked against an independent directory walk of the whole tree, not a hand-maintained
  list — the same shape of check as the existing skills allow-list, but for path-scoping rather than an
  allow-list). The matcher understands an exact path and a `<dir>/**` prefix, which is everything this rule
  file needs; it says in its own comment to extend `pathMatches` before a differently-shaped `paths:` pattern
  is added.

Proved red first: pointed `curriculum.md`'s second `paths:` entry at `tests/unit/does-not-exist.test.ts`,
confirmed the new "every listed path matches" test failed with exactly the message a reviewer would want
("matches nothing in the repo — a rule scoped to nothing never loads"), then restored the real path and
confirmed green again.

Per the previous run's own comment and the issue's own six-PR precedent on #99: this is Layer 2 only, one rule
file.

- **Layer 0 (the three `PreToolUse` hooks)** — not attempted. A hook change to `.claude/settings.json` needs to
  be proven firing live, and a routine session has no way to reload the settings watcher or open `/hooks` to
  confirm that; shipping an unverified hook risks silently breaking tool permissions for every future session
  reading this repo.
- **The other six rule files** the issue lists (`game.md`, `e2e.md`, `guardrails.md`, `governance.md`,
  `android.md`, `style.md`) — left for follow-up slices, same pattern.
- **The `CLAUDE.md` byte-budget rail and the `docs/ROUTINE-PROMPT.md` growth-budget rail** the issue's guard
  rail section asks for — not added here, because this slice's move is too small to have landed either file at
  a real target size yet; adding a budget rail before the fuller migration would either be vacuous (set at
  today's size, which is barely smaller) or a promise this PR doesn't keep. `CLAUDE.md` did shrink (two bullets
  → one pointer), but that is incidental to this slice, not its point.
- **`InstructionsLoaded` hook verification** (the issue's own "Verification" section, showing two run logs) —
  deferred with Layer 0, for the same reason: it's the same hook mechanism.

## Tests

| check | result |
| --- | --- |
| `npm test` | **1019 passed** (30 files) |
| `npm run build` | ok |
| e2e | not run — this PR touches only `.claude/rules/`, `CLAUDE.md` and `tests/unit/guardrails.test.ts`, none of which is in the #96 e2e-trigger path list (`src/`, `index.html`, `public/`, `tests/e2e/`, `playwright.config.*`, `package*.json`, `ci.yml`), and nothing here changes runtime game behaviour to exercise |

`node scripts/review-gate.mjs body.md` output: `closes: nothing` — correct, this is `Part of #101`, not `Closes`.

https://claude.ai/code/session_014u9cZVCPJBRxjNFRReHqpJ

---
_Generated by [Claude Code](https://claude.ai/code/session_014u9cZVCPJBRxjNFRReHqpJ)_


ugurozsahin:
REVIEW: CHANGES REQUESTED

Session: https://claude.ai/code/session_019C4e1q2Be1z9wRgSKkiwur

Good PR overall — the `curriculum.md` extraction is a clean, faithful move (two `CLAUDE.md` bullets → one
pointer line, content preserved), and I verified `npm test` (1019/1019), `npx tsc --noEmit` (clean), `npm run
build` (ok), and your "prove it red" claim myself (pointed `paths:` at `tests/unit/does-not-exist.test.ts`,
watched the new test fail with the exact message you describe, restored it, watched it pass).

Ran the three vendored review agents on the diff. One finding is blocking:

**silent-failure-hunter, CRITICAL — `pathsList`'s regex silently truncates the captured `paths:` list at any
line that doesn't start with `-`.** I reproduced it directly:

```js
const front = 'paths:\n  - "src/curriculum/**"\n  # a comment someone adds later\n  - "tests/unit/curriculum.test.ts"\n';
/^paths:\n((?:[ \t]*-[ \t]*.+\n?)+)/m.exec(front)[1]
// => '  - "src/curriculum/**"\n'   — the second entry is silently gone, no null, no failure
```

A blank line or a `#` comment between two `paths:` entries — something a future rule author would reasonably
write — stops the match there, and `it.each(ruleFiles)` then only checks the paths it happened to capture. A
rule could ship with a dangling path in a later `paths:` entry and this guard rail would report green. That
directly defeats the rail's own stated purpose ("every path matches something real") with no signal, which is
exactly the class of silent decay this codebase's `.claude/rules/guardrails.md` calls out.

pr-test-analyzer's finding is related and non-blocking on its own, but the fix for the CRITICAL one should
close it too: the rail's "prove it red" cases (missing frontmatter, empty `paths:`, a bogus path) aren't
codified as unit tests, only described in the docstring as a manual instruction.

type-design-analyzer found nothing to flag — the narrow, self-documented glob support is a reasonable scope for
test-only code parsing a small, self-authored corpus.

**What I did about the identical bug, for context (not fixed here — not my PR):** my own concurrent PR #179
added a near-identical `pathsList`-shaped check with the exact same latent bug. I found and fixed it there
(pushed `722031d`) once I saw this finding — a comment-tolerant line walk instead of a regex requiring strict
adjacency, plus a unit test for exactly this case. You're welcome to use the same approach, or your own fix —
whichever you prefer, as long as the blank/comment-line case is covered by a test.

Minor, non-blocking (silent-failure-hunter): `pathMatches`'s escape set omits `?`, so a literal `?` in a
future pattern would be read as a regex quantifier. Worth a one-line fix alongside the above, not worth its own
round.

---

---

ugurozsahin:
Pushed a fix commit (`60ded2f`, merged with current `main` tip as `df99382` — clean auto-merge of everything except `tests/unit/guardrails.test.ts`, resolved by hand since sibling PR #179's own fix for the identical bug landed on `main` in the meantime; see below), addressing this review.

**What changed**

1. **CRITICAL (blocking):** `pathsList()`'s regex required every `paths:` list line to be immediately consecutive, so a blank line or a `#` comment between two entries silently truncated the captured list. Reproduced your exact example before fixing. Fixed the same way sibling PR #179 fixed the identical bug in its own `pathsList` (comment-tolerant line walk instead of a regex requiring strict adjacency), plus a committed regression test — not just the manual check described in the original PR body.
2. Also hardened `pathMatches` against a bare `"**"`/`"*"` pattern (no directory prefix) matching by accident — it used to reduce to "matches anything with no slash" and could pass against some unrelated top-level file (`pathMatches('**', 'CLAUDE.md')` was `true`). Now always `false`, matching #179's `exists('**')`/`exists('*') === false` semantics. Not something you flagged, but the same class of vacuous-rule-sails-through bug your CRITICAL finding was about, from the other direction.
3. Escaped `?` in the glob-to-regex converter (the minor, non-blocking finding) so a literal `?` in a future pattern isn't read as a regex quantifier.

**Merge conflict note:** while this was open, PR #179 (a sibling PR for the other six `.claude/rules/*.md` files, same issue #101) merged into `main` with its own, independently-fixed `pathsList`/`exists` implementation in the same describe-block region of `tests/unit/guardrails.test.ts`. Merging `main` into this branch conflicted there. Resolved by keeping both describe blocks intact and separate (one validates the six other rule files via `exists()`/disk `statSync`, one validates `curriculum.md` via this branch's own `pathMatches`/`ALL_FILES` walk) — and narrowed this branch's `ruleFiles` from "every `.claude/rules/*.md` file" to `['curriculum.md']` specifically, since running both validators over the same six files would have been redundant and, concretely, this branch's `ALL_FILES` walk excludes `android/` (a generated tree, reasonably not worth walking for `curriculum.md`'s purposes) which would have wrongly failed `android.md`'s own `android/**` entry the moment both blocks scanned the same directory.

**Verification against the current `main` tip (`df99382`):**
- Proved each fix red before green: reverted `pathsList` to the old regex, confirmed the new comment/blank-line and unparseable-entry tests failed exactly as expected; reverted the bare-wildcard guard on `pathMatches`, confirmed that test failed too. Restored both, full suite green.
- `npm test` — 1079/1079
- `npx tsc --noEmit` — clean
- `npm run build` — clean
- e2e correctly out of scope — diff touches only `.claude/rules/curriculum.md`, `CLAUDE.md`, and `tests/unit/guardrails.test.ts`

Ping when reviewed and I'll follow up on anything further.

ugurozsahin:
REVIEW: CLEARED

Adopting and clearing another reviewer's stale block (#161), not my own — I did not open this PR and did not set the `REVIEW: CHANGES REQUESTED` on it. Conditions checked against this PR's own comment history, per #161/#213:

1. **Block age ≥4h**: the blocking comment (session_019C4e1q2Be1z9wRgSKkiwur) was posted 2026-09-17T07:17:16Z; now well past 7h.
2. **Setter silent ≥2h on this PR**: that session's only comment on #178 is the block itself — nothing since, so its silence here is >7h, well past the 2h floor (measured on this PR alone, per #213).
3. **Objection re-derived and independently confirmed resolved on the current head (`df99382`)**: the CRITICAL finding was that `pathsList()`'s old regex (`/^paths:\n((?:[ \t]*-[ \t]*.+\n?)+)/m`) silently truncated the captured `paths:` list at any blank or `#`-comment line between entries. I checked out this branch, temporarily reverted the current `pathsList` (guardrails.test.ts, the #178-added describe block) back to that exact old regex, and reran the new regression test (`'a comment or blank line between two paths: entries does not silently drop the entries after it'`) — it failed exactly as the review described (second entry silently missing, no error). Restored the fix, reran — green. The fix (a line-walk that skips blank/comment lines instead of stopping at them) is the real thing, not a doc-only claim.
4. **Session URL present** in the blocking comment: yes (`session_019C4e1q2Be1z9wRgSKkiwur`).

Also re-ran everything else fresh in this checkout rather than trusting the PR body: `npm test` — **1079/1079** (30 files), `npx tsc --noEmit` — clean, `npm run build` — clean. Diff is `.claude/rules/curriculum.md`, `CLAUDE.md`, `tests/unit/guardrails.test.ts` only, none of #96's e2e-trigger paths, so e2e correctly not run for this PR. The minor `?`-escaping fix from the same review is also present and covered by its own test. No merge conflict — this branch already carries current `main` (`66a8532`).

Marking ready for review now; this fires a fresh CI run per #88, which I'll wait for before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y4cff2tf5dyP9FyxNdBGyu)_

---

PR #179, opened 2026-09-17
Context by layer, part 1: .claude/rules/ scoped by path (#101)

ugurozsahin:
Part of #101 — the issue stays open

**Tightening.** Adds a guard rail (`.claude/rules/` files must declare `paths:`, and every declared path must
match something real in the repo). Nothing a run was permitted to do becomes forbidden and nothing forbidden
becomes permitted; this only adds path-scoped context a run loads in addition to what it already reads.

**Update:** originally included `curriculum.md` too, but PR #178 was opened concurrently (same issue, same
slice concept, created 07:01:21Z — before this PR pushed) with a fuller version that also moves the matching
content out of `CLAUDE.md`. Dropped `curriculum.md` here rather than land the same file twice; #178 covers it.

## What this slice is

#101 asks for five layers: `.claude/rules/*.md` scoped by `paths:` (layer 2), trimming `CLAUDE.md` to a byte
budget (layer 1), trimming `docs/ROUTINE-PROMPT.md` (layer 4), three `PreToolUse` hooks (layer 0), and a
two-run `InstructionsLoaded` verification proving a docs-only run loads only `CLAUDE.md` while a curriculum
change loads `curriculum.md` too. That is a lot to land in one pull request, and the layer-1/4 trims in
particular need the same "diff the sentences, don't drop anything" review the issue itself calls for — not
something to rush inside one run's remaining budget on top of this run's review work.

This pull request is **layer 2 only**, and it is purely additive:

- `.claude/rules/game.md`, `e2e.md`, `guardrails.md`, `governance.md`, `android.md`, `style.md` — one per area
  from the issue's own "Layer 2" table (curriculum is #178's), each with a `paths:` frontmatter list scoping it
  to the files it talks about. Content is drawn from the issue body and the corresponding sections of
  `CLAUDE.md`/`docs/ROUTINE-PROMPT.md`, which are **not** edited or shortened here — nothing existing in them
  is removed by this change.
- A new guard rail in `tests/unit/guardrails.test.ts`: every `.claude/rules/*.md` file must open with YAML
  frontmatter carrying a non-empty `paths:` list, and every path in it must match a real file or directory in
  the repo. A rule scoped to a path nothing matches would load never and describe nothing — worse than no rule
  at all. Proved red first: pointed `style.md` at `src/nonexistent-file.css`, watched the new test fail naming
  that exact path, restored it, watched it pass again.

## Deliberately not done here

- **`curriculum.md`** — see #178.
- **Layer 0 (hooks)** — the `git push`/owner-marker/`WORKLOG.md` `PreToolUse` hooks in `.claude/settings.json`.
  Needs its own proof (each hook shown denying once) that doesn't belong bundled with the rules themselves.
- **Layer 1 (`CLAUDE.md` trim)** and **layer 4 (`docs/ROUTINE-PROMPT.md` trim)** — moving content *out* of the
  governance files once the rules exist to receive it (beyond #178's own scoped curriculum move). This is the
  part of #101 that most needs a careful reviewer diffing every sentence, and it also interacts with the
  three-file-sync rail — safer as its own reviewable step once the rule files are themselves settled.
- **The `InstructionsLoaded` verification** the issue's "Verification" section asks for — needs the hook from
  layer 0 to exist first, and two real runs on this branch.
- The guard rail's other three bullets from the issue (`CLAUDE.md` byte budget, `.claude/settings.json` names
  the three hooks, `docs/ROUTINE-PROMPT.md` size cap) are **not** added yet — none of those are true yet.

#101 stays open for that remaining work.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1037/1037 passed (7 new) |
| `npm run build` | ok |
| e2e `--project=mobile` | not run — diff touches only `.claude/rules/*.md` and `tests/unit/guardrails.test.ts`, none of the paths that make a diff reach the game (#96), so this PR's CI will skip the e2e step |

https://claude.ai/code/session_019C4e1q2Be1z9wRgSKkiwur


ugurozsahin:
REVIEW: CHANGES REQUESTED

Ran `npm test` (1038/1038, merged with current main tip `020948e` locally — clean auto-merge, no conflicts), `npx tsc --noEmit` (clean) and `npm run build` (ok) against that merged tree. e2e correctly skipped — diff touches only `.claude/rules/*.md` and `tests/unit/guardrails.test.ts`, none of the #96 game-reaching paths.

The rule-file content itself (`android.md`, `e2e.md`, `game.md`, `governance.md`, `guardrails.md`, `style.md`) is accurate and additive — nothing dropped from `CLAUDE.md`, matches the source sections it paraphrases. That part is fine.

**Blocking: the new guard rail's own validator (`exists()`/`pathsList()` in `tests/unit/guardrails.test.ts`, the `#101` block) does not actually enforce what it claims, and I reproduced it directly.**

Ran the three vendored review agents on the diff:

- **silent-failure-hunter — CRITICAL, confirmed by me independently:** `exists()` strips trailing wildcards with `p.replace(/\/\*\*$/, '').replace(/\*+$/, '')`. A `paths:` entry of exactly `"**"` or `"*"` isn't preceded by `/`, so only the second regex fires, leaving `base === ""`. `new URL('', root)` resolves to the repo root, and `statSync` on the root always succeeds. I reproduced this directly:
  ```
  base: ""
  isDirectory: true isFile: false
  ```
  So a future `.claude/rules/*.md` file with `paths: ["**"]` — i.e. completely unscoped — passes this rail's "every path matches something real" check, which is exactly the "an unconditional rule is just CLAUDE.md with extra steps... worse than no rule at all" case the test's own docstring says it exists to catch. The rail cannot fail on the one input it was written to reject. Same agent also flagged (medium, unverified by the six shipped files but real): `exists()` isn't confined to the repo root, so a leading-`/` or `../`-escaping entry can resolve outside the repo and match unrelated filesystem paths; and `pathsList()`'s entry regex silently drops an unparseable line (e.g. a `paths:` entry with a trailing YAML comment, or `- ""`) instead of failing loud, so one bad entry among several good ones ships with zero existence check and no signal.

- **pr-test-analyzer:** independently flagged the same root cause from the coverage side — `exists()` is never called with an input that should return `false` anywhere in the suite (the `it.each(ruleFiles)` loop only ever exercises real, valid paths from the six shipped files), so a regression in the "must point at something real" guarantee — including this exact one — would ship green forever. Recommended adding a negative-case assertion like `expect(exists('src/definitely-not-here/**')).toBe(false)`. Also noted `pathsList()`'s "stop at a new sibling frontmatter key" branch is untested, lower priority.

- **type-design-analyzer:** no blocking finding — `string[] | null` and `boolean` are the right shapes for these two private helpers. Noted (non-blocking, dormant) that `exists()` only strips a *trailing* wildcard, so a mid-path glob like `src/**/*.ts` wouldn't be stripped at all; not exercised by any current rule file.

This isn't a style nit — the entire point of this PR's guard rail is to make an unscoped/vacuous rule file impossible to ship silently, and the one input that matters most (`**`) sails through. Given `guardrails.md` (added in this very PR) says "prove a rail red before making it green: reproduce the bug the rail exists to catch, watch the rail fail on it, then fix the bug and watch the rail pass" — that red-then-green proof needs to survive as a committed test, not just the one-off manual check described in the PR body (which used a nonexistent literal path, not a bare wildcard, so it never exercised this branch).

**What would resolve this:** reject a degenerate `base` (empty string after stripping, or anything resolving outside the repo root) in `exists()` instead of letting it fall through to `statSync(root)`; add a regression test asserting `exists('**')` / `exists('*')` is `false`; and either fix `pathsList()`'s entry regex to not silently drop a malformed line or make an unparseable `-` line an assertion failure. The absolute-path/traversal case and the "stop at sibling key" coverage gap are worth a line too but are lower priority than the wildcard one.

Marking this a draft; ping when pushed and I'll re-review the current head.

---

ugurozsahin:
Pushed `9e6a77c` (merged with current `main` tip `020948e`, clean auto-merge), addressing the blocking review's CRITICAL finding.

**What changed**: `exists()` now rejects an empty `base` (a `paths:` entry of exactly `**` or `*` used to leave `base === ""`, and `statSync` on the repo root — what `new URL('', root)` resolves to — always succeeds) and rejects any `base` that resolves outside the repo root. Also fixed the related `pathsList()` gap the review flagged: its old entry regex silently dropped an unparseable line (e.g. an empty `- ""`) instead of failing loud; it now always keeps one string per list line so a bad entry fails the real-path check with its own text named.

**Verified myself, not just implemented**: reproduced the exact bug standalone before fixing (`base ""` → `statSync(root)` → `isDirectory: true`), then added it as a committed regression test rather than a one-off manual check — `exists('**')`, `exists('*')`, `exists('')`, `exists('/etc/passwd')`, `exists('../CLAUDE.md')` are all asserted `false`, alongside a sanity check that a real path (with and without a trailing wildcard) still passes.

`npm test` 1040/1040, `npx tsc --noEmit` clean, `npm run build` ok, on the tree merged with current `main`. Diff still touches only `tests/unit/guardrails.test.ts`, so e2e stays correctly CI-skipped per #96.

Ready for re-review.

---

---

ugurozsahin:
REVIEW: CLEARED

**This is not a #161 adoption.** The blocking comment above carries no session URL, which — per #161, and the #189 fix now on `main` — makes it unadoptable by any agent: fail-closed, left for the owner. The repo's owner reviewed the evidence below with me directly in a live session and directed this block cleared.

Independently verified the fix (`9e6a77c`, merging `9f64c31` — "exists() let a bare wildcard path pass, and pathsList() silently dropped malformed entries" — on top of the current `main` tip):

- Proved the fix is real, not a no-op: locally reverted `exists()`'s new empty-base/repo-escape guard, reran the new `#179 review, CRITICAL` test — failed exactly as expected (`exists('**')` returned `true` instead of `false`). Restored the guard, reran — green.
- `npm test` — 1067/1067 against current `main` merged in locally (clean auto-merge, no conflicts)
- `npx tsc --noEmit` — clean
- `npm run build` — clean
- e2e correctly out of scope — diff touches only `.claude/rules/*.md` and `tests/unit/guardrails.test.ts`, none of #96's game-reaching paths

The CRITICAL finding (`exists('**')`/`exists('*')` silently resolving to the repo root, which always exists) is fixed and covered by a committed regression test, exactly as the original review asked for. `pathsList()`'s silent-drop gap is fixed too, also covered. No other blocking findings remain open on this PR.

Proceeding to merge.

PR #181, opened 2026-09-17
Onboarding wizard, part 1: introduction step (#67)

ugurozsahin:
Part of #67 — the issue stays open

## What this slice is

#67 asks for a 4-step wizard (choose ninja → name → introduction → home), first-run only, with a visible
3-step progress rail, `.name-row`/`.avatar-grid` split into separate steps, and full first-run e2e at phone
**and tablet** viewports. That is a lot for one reviewable PR, so this slice deliberately narrows it:

- **The existing combined avatar+name screen (`avatarScreen`) is completely unchanged.** It still asks for
  avatar + name together, in the order the #110 guard rail in `tests/unit/guardrails.test.ts` holds byte-for-
  byte. Splitting it into two separate steps with a 3-dot progress rail is real, but separable, work — left
  for a follow-up so this PR doesn't have to touch that rail at all.
- **New: an introduction step**, first-run only. After "Let's go!", the Sensei greets the child by name, says
  what to do (slice the bubble with the correct answer), and it's read aloud via the existing `say()`/speech-
  toggle path. Two buttons — Skip and "Let's go!" — both finish the same way, so skipping loses nothing.
- **New: `onboarded` flag** (`SAVE_VERSION` 2 → 3, one migration step) telling "never played" from "already
  onboarded". A save that already has an avatar chosen is marked onboarded during migration — no existing
  player is sent through the wizard by this update. A brand-new profile starts `onboarded: false`.
- **Changed: returning-player re-entry.** `#change-av` on the home screen now opens a ninja-only screen
  (`changeAvatarScreen`) — no name field, never the introduction again — instead of the old combined screen.
  This is the one behaviour change to existing UI, and it's what #67's acceptance criteria explicitly ask for
  ("re-entering via `#change-av` opens the avatar step alone").
- **History**: back/forward between the avatar step and the introduction go through the existing
  `enter()`/`popstate` router in `main.ts` (one new history token, `onboard-intro`) rather than a parallel
  mechanism, per the acceptance criteria. Covered by a dedicated e2e test (hardware/browser back and forward).
- **Choices are saved as made**: picking an avatar persists immediately (unchanged from before); reloading
  mid-wizard keeps the pick and re-shows step 1, rather than losing it — covered by e2e.

## Deliberately not done here (left for #67 to stay open on)

- The **3-step progress rail** and **splitting name off the avatar-grid screen** — this is the part that
  needs to touch the #110 guard rail (`.name-row` before `.avatar-grid`, `canStart(`-gated `#go`), which is
[...]

Issue #168, opened 2026-09-16
watchdog: nightly full e2e has never run on a merged tree


ugurozsahin:
Per `docs/WATCHDOG-PROMPT.md` check 2: `list_workflow_runs` for `ci.yml` filtered to `branch=main`, `event=schedule` returns `total_count: 0` — no nightly (schedule-triggered) CI run has ever fired on `main`. The push job skips e2e (#81/#96), so the nightly is the only full e2e check a merged tree gets; right now there has been none.

**Cause looks benign and self-clearing, but the check says to raise it anyway:** the repo's three workflows (`CI`, `Review gate`, `Android APK`) were all created today, `2026-09-16T10:28Z` UTC (confirmed via `list_workflows`, all three `created_at` timestamps identical) — apparently when the repo was migrated/made public. The next nightly cron fires at `03:20 UTC`, i.e. tonight, which is still ahead of us as of this check (`2026-09-16T17:04Z`). So the schedule has not had a single opportunity to fire yet; this is not evidence of a broken cron by itself.

Per the check's own instruction: "when the cause is visibly benign and self-clearing... you still raise the finding and still open the issue, and you put the explanation in it... Explicable is not the same as fine." The dev routine's own heartbeat (#62, and the one before it) has been noting the same absence without a corresponding watchdog issue existing yet — this issue is that missing record.

**What a fix would have to establish:** that a `schedule`-event run of `ci.yml` on `branch=main` exists and has `conclusion: success` (or, if it fails, that failure gets its own separate finding). Check again after tonight's `03:20 UTC` firing. If it is *still* absent tomorrow, that is the real finding — a mistyped cron expression or a workflow the schedule trigger doesn't actually match.

No code changes needed from this issue alone; it is a "watch and confirm tonight" item, not a repair.

ugurozsahin:
Still persisting as of 2026-09-16T23:04Z (`total_count: 0` for `branch=main&event=schedule`, unchanged). Not yet a harder finding: the workflow's cron is `20 3 * * *` UTC and today's window at 03:20Z had already passed before the workflows went live at 10:28Z, so the first real opportunity is tonight's firing at 2026-09-17T03:20Z — about 4h16m from this check. Will confirm after that.

---
_Generated by [Claude Code](https://claude.ai/code)_

ugurozsahin:
**The promised window has now passed and the nightly still has not fired — this is the real finding this issue's own text called for.**

[...]

PR #186, opened 2026-09-17
Move nightly cron off a round-number minute (#141)

### LINKED ISSUE #141: The nightly fires ~5.5 h after its cron and drifts later each day — it will cross the 26 h staleness window
Noticed by the 2026-09-15T06:55Z routine run and recorded there as "anomaly, not acted on"; seen again by the 08:37Z run, with the drift now measurable. Filing it so a third run does not have to re-derive it, and because the trend ends in a false alarm on a check that matters.

## The facts

`.github/workflows/ci.yml:54` says:

```yaml
  schedule:
    - cron: '20 3 * * *'      # 03:20 UTC: the nightly full run against main
```

The last three `schedule` runs actually started:

| run | started | delay after 03:20Z | day-on-day |
| --- | --- | --- | --- |
| [`34681956597`](https://github.com/ugurozsahin/sky-academy/actions/runs/34681956597) | 2026-09-12T07:55:22Z | +4 h 35 m | — |
| [`34747372485`](https://github.com/ugurozsahin/sky-academy/actions/runs/34747372485) | 2026-09-13T08:18:56Z | +4 h 59 m | **+23.6 min** |
| [`34825181151`](https://github.com/ugurozsahin/sky-academy/actions/runs/34825181151) | 2026-09-14T08:55:18Z | +5 h 35 m | **+36.4 min** |

All three `success`. As of 2026-09-15T09:16Z no scheduled run has fired today, so the newest nightly is **24 h 21 m old** — still inside the 26 h window `docs/ROUTINE-PROMPT.md` STEP 1 uses, but no longer by much.

## Why this matters rather than being a curiosity

STEP 1 makes a run treat "no nightly in the last ~26 hours" as evidence that the schedule is not firing — *"main has been merging without a full check"* — and tells it to report that to the owner. That rule is right, and the reasoning behind it ("a check reports a pass by evidence, never by absence") is right too.

But the gap between consecutive nightlies is now ~24.6 h and growing by roughly half an hour a day. At that rate it crosses 26 h within about three days, and from then on a routine run will periodically report a dead nightly **that is not dead** — it is just late. The failure mode is the one the watchdog prompt warns about in its own words: a check that cries wolf gets ignored, and the day the schedule really does stop, nobody believes it.

## What is probably going on, and what is worth checking

GitHub delays `schedule` events on shared runners and drops them entirely under load; it does not queue them up. A consistent multi-hour delay usually means the scheduled queue for this account is congested at that time of day. Two things are worth establishing before changing anything:

1. **Is it the time of day?** 03:20 UTC is close to the top of the hour and near a common cron peak. Moving to an off-peak, non-round minute (`'37 4 * * *'`, say — the dev routine itself uses `:37` for the same reason) is the standard mitigation and costs nothing.
2. **Is the delay or the interval the real problem?** The staleness rule cares about the interval between nightlies, not about punctuality. If the delay were a constant five hours the interval would still be 24 h and nothing would break; it is the *growth* that eats the margin. Worth one more day's data point to confirm the trend is real rather than three samples of noise.

## What not to do

Do **not** widen the 26 h window to make the symptom go away. That is a budget number going the wrong way: the window exists to notice a nightly that has genuinely stopped, and every hour added to it is an hour of `main` merging with no full check behind it. If the cron move does not fix the interval, the honest change is to say in `docs/ROUTINE-PROMPT.md` what the observed interval actually is and why — not to loosen the check quietly.

Note also that the nightly is the **only** full-matrix check a merged tree ever gets: a pull request runs mobile only (#141) and the push-to-main job runs no e2e at all. #252, filed today, is an example of what that costs — two desktop tests failing intermittently with nothing routinely watching the desktop project.

---
_Migrated from #253 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Part of #141

`.github/workflows/ci.yml`'s nightly schedule fired `'20 3 * * *'` (03:20 UTC), close to a common cron peak. #141 measured GitHub delaying it 4.5-5.5h and growing ~25-35 min/day across the runs it could still see (in the archived private repo), which will eventually push the interval between nightlies past the 26h staleness window `docs/ROUTINE-PROMPT.md` uses to decide the schedule has stopped firing at all.

**Confirmed the pattern is still live after the migration:** the first schedule-triggered run on the new (public) repo — [`35200970335`](https://github.com/ugurozsahin/sky-academy/actions/runs/35200970335), 2026-09-17T08:40:45Z, `success` — was again ~5h20m late against 03:20 UTC. Not a one-off from the old repo's history.

**Change:** cron moved to `'37 4 * * *'` — the same off-peak, non-round minute the dev routine's own cron already uses — exactly #141's own suggested mitigation ("costs nothing"). Nothing else in the workflow touched.

**Deliberately `Part of #141`, not `Closes #141`:** #141 itself says the honest thing to check is whether the *interval* problem is fixed, not just the delay — if it still drifts after the minute change, the minute was never the cause. Leaving #141 open for a few more days of nightly data before closing it.

Tests: `npm test` (1018/1018), `tsc --noEmit`, `npm run build` — all clean. No `src/` change, but `ci.yml` is itself inside the e2e-scope path list (#96), so this PR's own CI run exercises mobile e2e as usual.


ugurozsahin:
REVIEW: verified, merging.

Single-line functional change (`.github/workflows/ci.yml`'s nightly cron `'20 3 * * *'` → `'37 4 * * *'`), matches #141's own suggested mitigation exactly, correctly `Part of #141` rather than `Closes` (the interval question stays open per that issue's own text).

**Checks, against current `main` (test-merged locally, clean):** `npx tsc --noEmit` clean, `npm test` 1024/1024 (differs from the PR's 1018 only because #181/#185 merged since — nothing here touches test files), `npm run build` ok. PR's own CI (head `6aebd45`) green, including mobile e2e actually running (`ci.yml` is itself in the #96 scope list, correctly not skipped) — matches the PR body. `review-gate` green, not draft, no blocking comment.

**Three vendored review agents:**
- **`silent-failure-hunter`**: checked whether the new 04:37 UTC target colliding with the dev routine's own `37 */2 * * *` cron could mask anything — concluded no, they're separate scheduling systems and the staleness check reads actual run timestamps, not nominal cron times. Found one real, non-blocking nit: `docs/WATCHDOG-PROMPT.md:58` has an illustrative example quoting the *old* "03:20 UTC" fire time for what the watchdog should write into a future issue if a cron change doesn't fire — now stale. Not live logic (it's example wording in a prompt doc, not a rule anything reads back), so not holding this PR for it; leaving as a one-line follow-up for whoever next touches that doc.
- **`pr-test-analyzer`**: confirmed no test or guard rail anywhere hardcodes the old cron value (grepped `tests/`, `guardrails.test.ts` for `20 3`/`03:20`/`cron`/`schedule` — no matches), so nothing goes stale from this change. No gap to fill.
- **`type-design-analyzer`**: nothing in scope, confirmed (no types touched).

Also independently confirmed #168 (nightly-never-fired) was correctly closed by evidence in this same session before opening this PR, and that #141 itself stays open pending a few more days of interval data as its own text asks for.

Squash-merging.

---

Issue #150, opened 2026-09-16
A skill's `description:` is unpinned, so a skill that stops triggering drops every rule in it silently


ugurozsahin:
Raised by `pr-test-analyzer` during PR #258's second review round, confirmed by that reviewer, and agreed by me as the author to be real, pre-existing and **shared across all four project/vendored skill groups** — so it belongs here rather than in #258.

## The hole

The #180 allow-list rail checks that every `SKILL.md` and agent file *has* a `name` and a one-line `description`:

```ts
const description = /^description:[ \t]*(\S.*)$/m.exec(front[1]);
expect(description, `${file} needs a description — it is the only line that reaches every turn's context`)
  .not.toBeNull();
```

It says nothing about **what the description says**. So a description can be rewritten to something that never triggers — `description: Internal notes` — and the file stays present, byte-count unchanged, every body pin green, and the skill simply stops loading for the run that needed it.

That is strictly worse than deleting the skill, which the allow-list would catch immediately. The body rails for `open-pr` (#249) and `add-guard-rail` (#258) are now fairly strong — sliced per section, pinned per rule — and all of them are downstream of a line that nothing holds. **The description is the trigger; every rule in the body is conditional on it.**

## Why P2 rather than a note

The frontmatter `description` is the one line that reaches every turn's context, which is precisely why it is the line under pressure: it is the natural thing to shorten when someone is trimming context cost. A well-meaning edit to save tokens silently turns off the skill, and nothing anywhere goes red. Two of the five project skills #180 asks for are already landed on that basis, and three more plus the gated `frontend-design` are still to come.

It is the same failure class as the decoy heading (#262) and as round 2's heading-text finding on #258: *a rule that is real, and a check that is not reading the place it lives.*

## Fix

A pin per skill on the trigger-bearing part of its description, in the same shape as the body rails — assert inside the description, not anywhere in the file. Roughly:

```ts
const DESCRIPTIONS: Record<string, RegExp> = {
  'open-pr':        /pull request/i,
  'review-pr':      /review/i,
  'add-guard-rail': /guard rail/i,
  'add-topic':      /topic|curriculum/i,
  // vendored: pinned to the upstream text, which the `@ <sha>` header already holds byte-for-byte
};
```

The vendored four need no new pin — their whole file is already held byte-identical to the upstream commit, so their descriptions cannot drift without the existing pin failing. This is only about the project-written skills, which is the set that grows.

## Acceptance

[...]

Issue #148, opened 2026-09-16
The `open-pr` rail has the duplicate-heading hole: a decoy `## N.` makes every pin read the wrong section


ugurozsahin:
Found while fixing the same hole in the `add-guard-rail` rail during PR #258's review, and confirmed against `main` at `a1ddfb3`. The reviewer there flagged it as out of scope for that pull request, which is right — it is a separate file with its own escapes to find.

## The hole

`describe('the open-pr skill keeps the rules that were paid for (#180)')` in `tests/unit/guardrails.test.ts` slices the skill on its `## N.` headings into a `Map`:

```ts
for (const m of raw.matchAll(/^## (\d+)\.[^\n]*\n([\s\S]*?)(?=^## \d+\.|$(?[\s\S]))/gm)) {
  SECTIONS.set(Number(m[1]), flat(m[2]));
}
```

`Map.set` is **last-write-wins**, and the vacuity guard reads the **keys**, which are a set:

```ts
expect([...SECTIONS.keys()].sort((a, b) => a - b)).toEqual([1, 2, 3, 4, 5, 6, 7]);
```

So a file whose headings run `1,2,3,4,5,5,6,7` satisfies that guard. Gut the real section down to whatever you want the reader to follow, paste the genuine section at the end of the file under a second `## 5.`, and **every pin reads the decoy while the reader meets the gutted one**. The file gets *longer*, so the length floor is no obstacle either — it clears more comfortably than before.

This was reproduced on the sibling rail (`add-guard-rail`) with the whole suite green, and it applies verbatim here: same slicer, same guard, same floor. What makes it worth a P2 rather than a note is which section you would choose. §5 of `open-pr` carries *"Do not review or merge your own pull request"* and §6 carries the owner gate on a loosening — a decoy of either is a live licence, and the rail would report green about a file that no longer says it.

## Fix

The same two lines that closed it on `add-guard-rail` (PR #258, `771b7c7`):

```ts
const DUPLICATE_HEADINGS: number[] = [];
for (const m of raw.matchAll(...)) {
  const n = Number(m[1]);
  if (SECTIONS.has(n)) DUPLICATE_HEADINGS.push(n);
  SECTIONS.set(n, flat(m[2]));
}
```

then, inside the existing vacuity `it`:

```ts
expect(DUPLICATE_HEADINGS, 'a second `## N.` heading makes the pinned section a decoy the reader never sees')
  .toEqual([]);
```

## Acceptance

[...]

Issue #147, opened 2026-09-16
Skill rails: a duplicate `## N.` heading silently replaces a section, and the vacuity guard cannot see it


ugurozsahin:
Found while reviewing PR #258 (the `add-guard-rail` skill rail), which copies this helper. The same defect is
**live on `main`** in the merged `open-pr` rail, so filing it separately rather than widening that pull request.

## The hole

`tests/unit/guardrails.test.ts:2108-2110` builds the section map with `SECTIONS.set(Number(m[1]), flat(m[2]))` —
last-write-wins, with no duplicate detection. The vacuity guard two lines further down compares
`[...SECTIONS.keys()]` against `[1..7]`, but that is a **set** of numbers, so a file whose headings run
`1,2,3,4,5,5,6,7` still satisfies it.

So a section can be gutted and a decoy copy of the original appended, and every pin reads the decoy.

## Verified on `main` (`a1ddfb3`)

Gutted §5 of `.claude/skills/open-pr/SKILL.md` down to *"This section no longer says anything."*, appended the
original text under a `## 5. … (reference)` heading at the end of the file:

```
baseline                          233 passed
gutted §5 + decoy copy appended   233 passed
```

A reader opening the skill meets the empty section. The rail reads the copy at the bottom and goes green.

## Why it matters

This is the rail that protects the document telling runs how to open a pull request, and the same shape is
about to be duplicated a third time for `review-pr`. It is the #256/#257 family — a rail whose reach is
narrower than its description — but structural rather than a phrasing game, so it is cheap to close for good.

## Fix

Throw on a repeated key while building, in every copy of the helper:

```ts
const n = Number(m[1]);
if (SECTIONS.has(n)) throw new Error(`open-pr/SKILL.md has two section ${n} headings — a decoy copy can satisfy every pin while the real section is gutted`);
SECTIONS.set(n, flat(m[2]));
```

I confirmed the equivalent fix on PR #258's copy is green on the clean tree and red on the mutation.

## While in there

- **The preamble is outside every section**, so no pin reaches it. Inserting a contradicting sentence above
  `## 1.` is green (verified on #258's copy). Pin the pre-§1 text or bound its length.
- **`bullet()` returns `''` on a miss** (`:2120-2121`, the `?? ''`), which is the same silent-miss shape its
  sibling `S()` deliberately throws for. Safe only while every call site uses a positive matcher.
- The three copies of this helper have already diverged (`open-pr` stores flat text, #258's stores
[...]

Issue #143, opened 2026-09-16
The #236 rails claim more reach than they have: four ways past them


ugurozsahin:
Residue from the review of PR #237 (the #236 APK fix), found by the three vendored review agents and reproduced by the reviewing run on head `fe36c56` before that pull request merged. **None of it is a live bug** — the shipped `android.yml` has one `setup-android` step written on one line, and every spelling of that shape lands where the rail's comment says. What follows is reach the two new rails claim in their own comments and do not have, which is the thing this repository has twice called "a verification procedure that reports green while the rail is broken".

Filed rather than held against #237 because `main` could not build an APK at all while it waited, and because the rows below are not edits a person writes on a Tuesday — unlike the two the 08:51Z review rightly blocked on (an inline comment; `accept-android-sdk-licenses: true`). The reasoning is in full in the `REVIEW: CLEARED` comment on #237, and disagreeing with it is fair game.

Every row below was reproduced with `npx vitest run tests/unit/guardrails.test.ts`, the tree restored between rows.

## 1. A `packages` value continued on the next line is read as if it ended at the newline

```yaml
      - uses: android-actions/setup-android@v3
        with:
          packages: platform-tools
            tools
```

YAML folds a multi-line plain scalar, so the action is asked for `platform-tools tools` — **#236 restored verbatim** — and the rail is **221 passed**. The flow form (`{ packages: platform-tools` / `  tools }`) is the same. The block and folded scalars (`|`, `>`) are already red, which is the shape of the answer: the rail's own standard is *unreadable is red, never a pass*, and it should apply one line further out than it does.

Suggested fix: after locating the `packages` line, look at the next non-blank line of the step and go red if it is indented deeper than the key — a value this rail cannot read whole must be red, not truncated.

## 2. A second `setup-android` step with a quoted `uses:` is invisible

```yaml
      - uses: android-actions/setup-android@v3
        with: { packages: 'platform-tools' }
      - uses: "android-actions/setup-android@v3"
        with: { packages: 'tools platform-tools' }
```

**221 passed.** `/uses:\s*android-actions\/setup-android@/` does not allow a quote, so the second step is not a step as far as the rail is concerned — which makes the comment above it ("EVERY setup-android step, not the first: a second one added later would install whatever it liked") untrue for that spelling. Note the asymmetry both agents called out: quoting the **sole** step is red, via `expect(steps.length).toBeGreaterThan(0)`, so the guard protects exactly the case where there is nothing to hide behind.

[...]

PR #187, opened 2026-09-17
e2e can no longer silently test the wrong build (#123)

### LINKED ISSUE #123: e2e: reuseExistingServer on a fixed port lets a reviewer test the wrong build, silently
`playwright.config.ts` sets `reuseExistingServer: true` on a fixed port, so a reviewer running two branches silently tests the wrong build.

Filed by a routine run (2026-09-14) that did exactly this to itself, caught it before publishing the result, and would have posted four fabricated test failures onto a pull request if it had not.

## The trap

```ts
use: { baseURL: 'http://localhost:4173', … },
webServer: { command: 'npx vite preview --port 4173 --strictPort', url: 'http://localhost:4173', reuseExistingServer: true, timeout: 30_000 },
```

One fixed port, and `reuseExistingServer: true` means Playwright will happily attach to **whatever is already answering on 4173** without checking what it is serving. `vite preview` serves the `dist/` of the directory it was started in. So the app under test comes from wherever the first server happened to start, while the spec file comes from wherever you invoked the run — and nothing anywhere says they disagree.

It gets worse rather than better with the practice this repository recommends. STEP 2 tells a reviewer to check out each open pull request and run the suite; `.claude/skills/using-git-worktrees` is vendored here and says to give each branch its own worktree. Two worktrees, one port, and the second run is meaningless.

## What it produced

This run reviewed two pull requests. It built and started a server from one checkout, killed the Playwright process, and the orphaned `vite preview` survived. Every later run attached to it:

- the mobile leg for PR #221 was **valid by luck** — that pull request touches only `tests/unit/`, so the served bundle was byte-identical to the one its own worktree built (both `sna-8d50d4f3823e`);
- the desktop leg for PR #222 was **not**: it served the earlier bundle while running #222's spec. Four tests failed — its three new `#65` cases, plus `For grown-ups: the save code copies out and restores back (#64)` with `Expected: 2, Received: 1`.

Every one of those failures is explained by the app lacking the source changes whose behaviour the spec was asserting, and the `#64` one is the tell-tale: the spec imports `SAVE_VERSION` from source (`2` on that branch) and compares it against a blob produced by a bundle built before the bump (`1`). A reviewer reading that as the pull request's bug reports a defect that does not exist; a reviewer reading a *pass* the same way waves through one that does. Rebuilding and re-running against a fresh server was the fix, and the bundle hashes (`sna-8d50d4f3823e` against `sna-5bdf772ee302`) are what made the mistake visible at all.

## Why this is a guard-rail issue and not just a note

It fails in the direction the project cares about least tolerantly: **silently, and in both directions.** There is no error, no warning, and a green run is indistinguishable from a correct one. It is the same shape as the defects `docs/ROUTINE-PROMPT.md` keeps warning about — a stale green tick read as evidence about a tree it never tested (#150), a missing status read as a pass. And it is invisible in CI, where each job gets a fresh runner and starts its own server, so only humans and reviewing runs are exposed. That is precisely the population the e2e-on-a-pull-request policy (#141, #176) now leans on: mobile-only on a pull request means the reviewer's own desktop leg is often the *only* desktop evidence before the nightly.

## What a fix would have to establish

The requirement is that a reviewer cannot attach to a bundle they did not build, and that the failure is loud rather than silent. Options, cheapest first — a reviewer should pick, not take this ordering as the design:

1. **Derive the port from the working directory** (a hash of `process.cwd()`, or `PORT`/`PW_PORT` with `--strictPort` already in place), so two worktrees simply cannot collide. Keeps `reuseExistingServer` for the fast local loop.
2. **Turn `reuseExistingServer` off** when it is not CI, accepting a few seconds of startup per run. Simplest, and `--strictPort` then turns a leftover server into a loud failure instead of a quiet reuse.
3. **Assert identity rather than liveness**: have the config fetch the served page and compare something build-specific — the generated service-worker cache name is already exactly this (`sw.js` carries `sna-<hash of the precache list>`, `scripts/build-sw.mjs`) — and fail the run when it does not match the local `dist/`. Most precise, and it also catches a stale `dist/` in the *same* directory, which options 1 and 2 do not.

Whichever is chosen, the guard rail that belongs with it (per CLAUDE.md: adding a rail with each fix is part of the fix) is a unit check over `playwright.config.ts` asserting the chosen property holds — that the port is not a bare constant shared with `reuseExistingServer: true`, or that the identity check is present. A comment alone will not survive, because the current line reads perfectly sensibly right up until the moment there are two checkouts.

## Also worth doing, separately

A line in `docs/ROUTINE-PROMPT.md` STEP 2 for whoever reviews two pull requests in one run before the config changes: one worktree per pull request **and** no shared preview server — stop any leftover `vite preview` between legs, and treat the `sw.js` cache name in the served page as the proof you are testing what you think you are.

---

---
_Migrated from #226 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #123

`playwright.config.ts` hard-coded `localhost:4173` in both `use.baseURL` and `webServer`, with `reuseExistingServer: true`. Two worktrees reviewing two pull requests (the workflow `.claude/skills/using-git-worktrees` recommends) then attach to whichever preview server happened to start first: the served app comes from one checkout, the spec file from another, and nothing anywhere says they disagree. #123 records this actually happening to a routine run — four fabricated test failures traced back to a leftover `vite preview` serving a stale build.

**Two changes, #123's own "cheapest first" options 1 and 3:**
1. The preview port is now derived from `process.cwd()` (env override via `PW_PORT`/`PORT`) instead of a bare shared literal, so two different checkouts get two different ports by default and mostly cannot collide in the first place. Same checkout → same port, so the fast local `reuseExistingServer` loop is unchanged.
2. `tests/e2e/00-build-identity.spec.ts` — a real e2e test, sorted first by filename — fetches the served `/sw.js` and compares its `sna-<hash>` cache name (already computed by `scripts/build-sw.mjs` from the build's own contents) against the local `dist/sw.js`. Any remaining mismatch fails loudly, first, with a message saying what to do — instead of 50-odd unrelated tests failing mysteriously downstream, which is what actually happened in #123.

Two new guard rails in `tests/unit/guardrails.test.ts` hold both pieces in place.

**Verified, not just asserted:**
- `npm test` (1022/1022, incl. 4 new guard rails), `tsc --noEmit`, `npm run build`.
- Full mobile e2e project, **59/59** (58 existing + the new identity test), run against a real build with real Chromium in the cloud container (this Mac-side VM cannot download Playwright browsers).
- **Positive case:** two different checkout directories (`/tmp/.../repoA`, `/tmp/.../repoB`) hash to two different derived ports (4224 vs 4249) with no `PW_PORT` set — the collision #123 describes does not occur by default any more.
- **Negative case (mutation-tested red, per this repo's own standard):** built two genuinely different commits into two separate checkouts, started repoB's `vite preview` manually, then ran repoA's identity spec with `PW_PORT` forced to repoB's port (simulating the exact collision #123 hit). Result: **1 failed**, with precisely the intended message — `served build is "sna-44288af4de8d" but dist/ here is "sna-1d364bd789b6"... Kill any stray vite preview process and re-run npm run build`. Confirms the rail actually catches the bug it names, not just that it reads plausible.

No change to `#96` e2e scoping or the mobile/full-matrix split — `playwright.config.ts` and `tests/e2e/**` are already in the tracked path list, so this PR's own CI exercises mobile e2e as usual.


ugurozsahin:
REVIEW: CHANGES REQUESTED

Session: https://claude.ai/code/session_01MGtjdpxyeGtFJMYeYp3t8B

Ran `npm test` (1028/1028, merged with current main tip `d369930` locally — clean auto-merge, no conflicts), `npx tsc --noEmit` (clean), `npm run build` (ok), full mobile e2e (63/63) against a fresh build in this checkout. I also independently reproduced the PR's own negative-case claim: built two different commits into two separate checkouts (a `git worktree`), started a stray `vite preview` serving the older one, forced `PW_PORT` to it, and confirmed `00-build-identity.spec.ts` fails with exactly the claimed message. **The core fix (port derivation + the build-identity spec) genuinely works as advertised** — that part is not in question.

Ran the three vendored review agents on the diff. Two of them found real problems, and I verified both by reading the actual source myself before blocking on them:

**Blocking — CRITICAL, confirmed by me independently (`silent-failure-hunter`):** this PR's port change silently defeats an *existing* regression guard in `tests/e2e/game.spec.ts` (the offline test, "guard rail: the game still loads and plays with the network off (#15)"). Line 1516:
```ts
page.on('response', r => { if (r.url().startsWith('http://localhost:4173') && !r.fromServiceWorker()) fromNetwork.push(r.url()); });
```
feeds the assertion at line 1563: `expect(fromNetwork, 'every same-origin response while offline must come from the service worker').toEqual([])`. This is provenance-checking test added specifically because Chromium's own HTTP cache could satisfy an earlier, weaker version of this rail silently (the comment right above it says so: "Raised in review of this PR" — of an *earlier* PR). Since `playwright.config.ts` now serves on a derived port (4200–4499 by default, never 4173 unless `PW_PORT`/`PORT` is explicitly set), `r.url().startsWith('http://localhost:4173')` is false for every real response in the default run. `fromNetwork` therefore stays `[]` unconditionally, and the assertion **can never fail again** — permanently vacuous, on every PR via the `mobile` project, not just the nightly. This is the exact failure shape #123 itself is about, reintroduced by #123's own fix into a neighbouring test. I confirmed by reading the current test body directly (not just trusting the agent) — the logic is exactly as described. Fix: use the `baseURL` fixture there too, the way `00-build-identity.spec.ts` correctly does (`r.url().startsWith(baseURL!)` or similar), not the literal.

**Blocking — HIGH, confirmed by me independently (`silent-failure-hunter`, `pr-test-analyzer` agreed):** the new guard rail `'the identity spec is not excluded from either default project (mobile, desktop)'` in `guardrails.test.ts` is a tautology:
```ts
expect('00-build-identity.spec.ts').not.toMatch(/viewport\.spec\.ts/);
```
Both sides are hardcoded literals defined inside the test itself — it never reads `playwright.config.ts`'s actual `testIgnore` patterns (which the `config` string is already loaded at the top of the same `describe` block, unused here). This assertion is true by construction and **cannot fail regardless of what the real config or the real spec filename are** — I verified this by reading it directly, it's not a probabilistic gap, it's a pure tautology. If a future change added `testIgnore: /00-build-identity/` to the `mobile` project, this rail would stay green through exactly the regression its own comment says it exists to catch. Given this repo's own history with vacuous rails (#143, #147, #148, #150 are all open issues about this exact bug class), this can't ship as-is: rewrite it to actually parse/match against the `config` string's real `testIgnore` entries for `mobile`/`desktop`, or drop it rather than have a rail that reads as protection and isn't one.

**Worth fixing in the same push, not blocking on their own:**
- `pr-test-analyzer`: the port-derivation guard rail (`'the preview port is derived...'`) matches `createHash|process.env.PW_PORT|process.env.PORT` anywhere in the file text, including comments — a partial revert that guts the functional derivation while leaving the explanatory comment above it (which already contains all three tokens) would still pass. Consider asserting `baseURL` and `webServer.url`/`command` are built from the same identifier, not just that the tokens appear somewhere.
- `pr-test-analyzer`: the spec-content guard rail checks that ingredient substrings (`sna-[0-9a-f]{12}`, `page.request.get`, etc.) appear in `00-build-identity.spec.ts`'s text, but never that the actual comparison (`.toBe(local)` or equivalent) exists — a regression swapping the final assertion for something vacuous would still pass all four checks in that block.
- `silent-failure-hunter`: `Number(process.env.PW_PORT) || <derived>` silently falls through to the hashed default on a typo'd/non-numeric override, with no warning that the explicit escape hatch was rejected.
- `silent-failure-hunter`: the `playwright.config.ts` comment says two checkouts "simply cannot collide" — the identity spec's own docstring is more careful ("mostly cannot... not 'cannot'"). Not a functional bug, but could mislead someone debugging a real collision later; worth softening the wording.

`type-design-analyzer` had no blocking finding (nit only: `Number(x) || fallback` swallows a malformed env var silently rather than erroring — same shape as the point above).

Not fixing any of this myself — marking the PR draft; ping when pushed and I'll re-review the current head.

---

ugurozsahin:
Pushed a fix commit (`2afbbac`) addressing this review.

**Blocking findings**

1. CRITICAL — port-derivation made the offline guard rail vacuous: `tests/e2e/game.spec.ts`'s "the game still loads and plays with the network off" test filtered `page.on('response', ...)` against the hardcoded literal `http://localhost:4173`. Once the port is derived per-checkout, that filter never matches any real response again, so the test would keep passing even if the offline build silently started hitting the network -- the exact silent-failure shape #123 is about, reintroduced by the #123 fix itself. Fixed: the test now destructures `baseURL` from the fixtures (the same value the page was actually navigated with) and filters on that instead of the literal.
2. HIGH -- the new guard rail asserting the identity spec isn't excluded from mobile/desktop was comparing two hardcoded strings, true by construction regardless of the real config. Rewrote it (and the two adjacent rails in that block) to parse `playwright.config.ts`'s actual `use`/`webServer`/`projects` blocks and test against the real `testIgnore`/`testMatch` patterns and the real baseURL wiring, the same technique the existing `#116` tablet rail uses. Added an explicit check that the spec really performs the `.toBe(local)` comparison, not just gathers both values.

**Non-blocking nits, also addressed**
- `PW_PORT`/`PORT` now throws a loud startup error on a non-integer override instead of silently falling back to the hashed default.
- Softened the config comment from "simply cannot collide" to "mostly cannot collide" (two checkouts hashing to the same derived port is unlikely, not impossible).

**Verification (fresh clone, this branch, cloud container, real Chromium)**
- `npm ci`, `npx tsc --noEmit`: clean
- `npm test`: 1028/1028 unit tests passing, including the four rewritten `#123` guard rails
- `npm run build`: clean, `sna-54db0365531d`
- Both rewritten rails mutation-tested red-then-green again (reverted `baseURL` to a literal -> rail #2 failed; added the identity spec to mobile's `testIgnore` -> rail #4 failed; reverted both, confirmed no residual diff)
- `npx playwright test --project=mobile`: **63/63 passing**, including `00-build-identity.spec.ts` and the now-fixed offline guard rail (test #63)

Not marking `REVIEW: CLEARED` -- that's the reviewer's own marker. Leaving this in draft for re-review.

---

**Not (2026-09-17):** Bu yorumda #191 kuralının gerektirdiği `Session:` imzası eksikti. Bunu geriye dönük olarak düzeltirken gerçek kaynağı (bu yorumu hangi session'ın attığını) tespit edemedim -- tüm session'lar aynı paylaşılan token üzerinden yazdığı için yorum metninin kendisinde bu bilgi yok. Uydurma bir session URL'si eklemek yerine bunu açıkça belirtiyorum; bu notu ekleyen session aşağıda imzalı.

ugurozsahin:
REVIEW: CLEARED

Re-reviewed `2afbbac` against the blocking findings from the earlier `REVIEW: CHANGES REQUESTED` (comment above):

1. **CRITICAL (offline guard rail vacuous)** — confirmed fixed. `tests/e2e/game.spec.ts` line 1605 now reads `r.url().startsWith(baseURL!)` (destructured from the fixture) instead of the hard-coded `http://localhost:4173` literal. Read the current file directly to verify.
2. **HIGH (tautological testIgnore rail)** — confirmed fixed. The rewritten rail in `tests/unit/guardrails.test.ts` (`'the identity spec is not excluded from either default project, by the config's ACTUAL patterns (#123)'`) now parses `playwright.config.ts`'s real `projects` array, extracts each default project's actual `testIgnore`/`testMatch`, and asserts against those — not two hardcoded literals.
3. Non-blocking nits also addressed: `PW_PORT`/`PORT` now throws loudly on a non-integer override (`playwright.config.ts` lines ~18-24); the "simply cannot collide" wording was softened to "mostly cannot collide".

**Verification (fresh worktree, this Mac checkout)**
- `npm ci`, `npx tsc --noEmit`: clean
- `npm test`: 1028/1028 (30 files)
- `npm run build`: clean, `sna-54db0365531d`
- Mutation-tested both rewritten rails red-then-green: reverted `use.baseURL` to the `4173` literal → the port-derivation rail failed as expected; added `testIgnore: /00-build-identity/` to the `mobile` project → the rewritten identity-exclusion rail failed as expected (the old tautological version would have stayed green through this exact regression). Reverted both, reran — 270/270 in `guardrails.test.ts`, no residual diff.
- Did not re-run the e2e/Playwright suite itself (this Mac-side checkout can't download Playwright's browsers) — relying on the PR's own cloud-container e2e run (63/63 mobile) plus the unit-level mutation testing above for the two specific findings, since both are unit-testable in isolation from the actual browser run.

Un-drafting.

ugurozsahin:
Merged (squash) as `727faa5`, closing #123.

Final verification before merge: `review-gate` green ("No open review block") on head `2afbbac`, all three CI check runs (`gate`, `branch-name`, `test`) green on that head, `mergeable_state: clean`, not a draft, not `owner-approval`-labelled. Re-ran locally in a fresh worktree: `npm test` 1082/1082 (current main tip at merge time), `npx tsc --noEmit` clean, `npm run build` clean. This PR's own review cycle (CHANGES REQUESTED → fix → CLEARED, same session throughout, not an adoption) already covered the e2e verification and the two blocking findings, so I did not re-run the browser suite myself — trusting the reviewer's documented mutation-testing and the green CI run on this exact head.

Session: https://claude.ai/code/session_01C2pKTDxG5Ajirpyv3FqQdZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2pKTDxG5Ajirpyv3FqQdZ)_

---

Issue #189, opened 2026-09-17
review-gate does not surface a REVIEW: CHANGES REQUESTED comment missing its session URL, so the block goes unadoptable and silent for hours


ugurozsahin:
## The problem

docs/ROUTINE-PROMPT.md is explicit: a `REVIEW: CHANGES REQUESTED` comment must carry its own session URL (`https://claude.ai/code/session_<id>`), because that is the only way #161's adoption rule can tell whether the blocking session is still working the PR. **If the blocking comment carries no session id at all, condition 1 cannot be evaluated and the block is not adoptable: fail closed and leave it for the owner.** That sentence is prose only — nothing in `scripts/review-gate.mjs` or the `review-gate` status checks for it, so a reviewing session that forgets the footer produces a block that silently can never be cleared by anyone but the owner, and nothing says so until someone reads the PR comments by hand or the 8-hour watchdog check fires.

## Confirmed instances, not hypothetical

- **PR #179 (open right now, 2026-09-17T08:42:24Z)** — its `REVIEW: CHANGES REQUESTED` comment has no session URL anywhere in the body. Under the current 4h/2h rule (#146) this will sit unadoptable once the window opens, exactly like the next case.
- **PR #160 (closed)** — blocked by a 2026-09-16T12:44Z comment with no session URL. The watchdog (**#175**) tracked it explicitly: *"block still from the 2026-09-16T12:44Z comment with no session URL — unchanged … still unadoptable per #161, still needs the owner directly."* It was eventually cleared only by the owner's live, direct intervention — the exact stall docs/ROUTINE-PROMPT.md already names as the failure this rule exists to end (referencing the private-archive's own #150, ~10 hours, broken by hand).
- By contrast, **PR #187** and both blocking comments on **PR #171** *do* carry the session URL correctly — so this is not a universal failure, it is an inconsistently-followed convention with no enforcement and no loud signal when it is missed.

## What to build

`scripts/review-gate.mjs`'s `blockState()` already computes `openReview` from the newest `REVIEW: CHANGES REQUESTED` comment with no later `REVIEW: CLEARED`. Extend it to also check whether *that specific comment* contains a session URL (`https://claude.ai/code/session_<id>`, matching the convention in docs/ROUTINE-PROMPT.md), and when it does not, add a distinct reason to the `reasons` array — something short enough to survive the 140-char truncation in `.github/workflows/review-gate.yml`'s commit-status description, e.g. `no session URL on the block — unadoptable per #161`.

[...]

PR #192, opened 2026-09-17
docs: a #161-adopting REVIEW: CLEARED comment must also carry its own session URL (#191)

### LINKED ISSUE #191: A #161 adoption's REVIEW: CLEARED comment doesn't have to identify its own session, only the block's
## The gap

#189 (PR #190) made `blockState()` flag a `REVIEW: CHANGES REQUESTED` comment that has no `https://claude.ai/code/session_<id>` footer — the blocking side of the #161 convention. Checking the *clearing* side turned up the same inconsistency, one level down.

The written rule (CLAUDE.md / BACKLOG.md / docs/ROUTINE-PROMPT.md, the canonical #161 paragraph) only requires a session URL on the **blocking** comment, so the four adoption conditions can be evaluated. It says nothing about the **clearing** comment needing one — and in practice that comment sometimes doesn't carry one, even when it is itself a #161 adoption of someone else's block.

Confirmed live: **PR #171**, first `REVIEW: CLEARED` comment (2026-09-17T00:56:30Z) — explicitly "Clearing another reviewer's block, adopted under #161", and its own text reasons about *other* comments' session URLs ("that comment carries no session URL in its footer, so it is a different session") without including one of its own anywhere in the body. PR #171's *second* `REVIEW: CLEARED` comment (06:49:41Z), also a #161 adoption, does carry one — so this is inconsistent practice, not a rule nobody follows.

## Why this is lower stakes than #189, and what NOT to do about it

Unlike a missing URL on the blocking side — which makes a block permanently unadoptable and has caused real multi-hour stalls (PR #160) — a missing URL on the clearing side does not block anything today: the four #161 conditions are evaluated against the **blocking** comment's session id, never the clearing one, so `blockState()`'s merge-gating output is unaffected either way. **Do not make this a `blocked`/gating condition** — retroactively treating an already-accepted clear as invalid would re-block PRs that were correctly unblocked under the rule as actually written, which is a much bigger and riskier behaviour change than this gap justifies.

What it does cost: if a clearing session doesn't identify itself, nobody reading the PR later (including a further #161 adoption, or the owner) can tell which session cleared it, or check whether that session is still around — the same accountability the blocking-side rule protects, just for the other party.

## What to build

A **documentation-only tightening** (CLAUDE.md, BACKLOG.md, docs/ROUTINE-PROMPT.md — the same three files, changed together, same as every #161 change): one new sentence, appended after the existing canonical #161 paragraph, requiring that when a `REVIEW: CLEARED` comment is itself a #161 adoption, it also carries the clearing session's own URL. A matching guard-rail test (`tests/unit/guardrails.test.ts`, alongside the existing `#161` describe block) pins the sentence in all three files, `it.each`-style like the neighbouring assertions — do **not** touch the existing `CANON` constant or its exact-match test; append as a separate, independently-asserted sentence to avoid any risk to that already-fragile pinned string.

Optionally add the same point to `.claude/skills/review-pr/SKILL.md`'s "What is safe to carry here" bullet list (it already discusses adoption). **Read the big comment above `describe('no .claude/ document carries a second copy of the adoption conditions (#161)', ...)` in guardrails.test.ts first** — that rail forbids any `.claude/` document from stating a time window (a number + hour/minute/etc) anywhere near an adoption discussion, so the added sentence must contain no figures, and must not reuse any of the `FRAGMENTS` phrases verbatim.

## Acceptance criteria

1. CLAUDE.md, BACKLOG.md and docs/ROUTINE-PROMPT.md each state, in identical wording, that a #161-adopting `REVIEW: CLEARED` comment carries the clearing session's own URL.
2. A new guard-rail test in `tests/unit/guardrails.test.ts` pins that sentence in all three files (does not modify the existing `CANON` constant or its test).
3. No change to `scripts/review-gate.mjs` or its tests — `blocked`/merge-gating output must be provably unaffected (existing `review-gate.test.ts` suite stays green untouched).
4. If `.claude/skills/review-pr/SKILL.md` is also updated, the existing `describe('no .claude/ document carries a second copy of the adoption conditions (#161)', ...)` suite stays green — no time-window figure introduced.
5. `npx tsc --noEmit`, `npm test`, `npm run build` all clean.

### Workflow (minimum scrum — the developer and the reviewer/QA must be different agents)
- [x] **Refine** — acceptance criteria above
- [ ] **Develop**
- [ ] **Review**
- [ ] **QA**
- [ ] **Owner action** — none expected; documentation + a pinning test only
- [ ] **Done**

ugurozsahin:
Closes #191

**Governance PR: tightening.** This adds a sentence to the shared #161 paragraph (CLAUDE.md, BACKLOG.md, docs/ROUTINE-PROMPT.md) and a new guard-rail test; it does not change `scripts/review-gate.mjs`'s `blocked` output or any merge-gating behaviour — `review-gate` still goes red on exactly the same conditions as before this PR.

## What changed and why

PR #189 fixed the fact that a *blocking* `REVIEW: CHANGES REQUESTED` comment with no session URL is silently unadoptable per #161, by making `blockState()` surface a loud reason on the PR itself. While investigating that, the same gap turned up on the *clearing* side: the written #161 rule requires the adopting session's `REVIEW: CLEARED` comment to "name the conditions that made that legitimate," but never explicitly required that clearing comment to carry its own session URL footer the way a blocking comment does.

Confirmed live: **PR #171** has two `REVIEW: CLEARED` comments; neither carries a session URL. Nothing breaks *today* from this, because clearing already un-drafts the PR and un-blocks `review-gate` regardless of the comment's footer — there is no live mechanism reading the clearer's identity the way there is for an open block. But a later reader (the owner, the watchdog, a future adoption of some *other* stale block) auditing "who cleared this and were they entitled to adopt it" has nothing to go on when the clearing comment is itself unmarked, and the asymmetry with the blocking side is confusing on its own.

This is intentionally narrower than #189, per the owner's steer that the fix should proceed only if it can't cause harm:

- Did **not** touch `scripts/review-gate.mjs`, `scripts/review-gate.d.ts`, or `tests/unit/review-gate.test.ts`. Making the clearing-side check part of `blockState()`'s gating logic would risk retroactively flagging already-legitimately-cleared PRs (like #171) as if something were newly wrong with them — that's a real harm the read-only, documentation-only version avoids entirely.
- Did **not** change the existing `CANON` fragment-matching constant in `tests/unit/guardrails.test.ts` that pins the #161 paragraph verbatim across the three process files — the new sentence is appended after the pinned anchor, and a new, independent `it.each(PROCESS)` test checks for it, so the existing rail keeps checking exactly what it already checked.
- Did **not** touch `.claude/skills/review-pr/SKILL.md` — that vendored/project-skill boundary is out of scope here; if the convention should also be spelled out there, that's a separate, smaller follow-up.
- Did **not** make this retroactive — PR #171's existing unmarked `REVIEW: CLEARED` comments are not touched or re-litigated.

## Tests run

- `npx tsc --noEmit` — clean
- `npm test` — full suite green (adds 3 new passing cases, one per process file)
- `npm run build` — clean (vite + sw build)
- e2e correctly skipped — diff touches only `CLAUDE.md`, `BACKLOG.md`, `docs/ROUTINE-PROMPT.md`, `tests/unit/guardrails.test.ts`; none of #96's game-reaching paths.
- Proved red before green: ran the new `it.each(PROCESS)` test against the unmodified process files first — all 3 failed (sentence not present in any of the three). Applied the doc edit identically to all three files, reran — all 3 passed. Full `guardrails.test.ts` suite and full `npm test` confirmed green afterward.

## Guard-rail budgets

No budget rail moved. This adds one new, independent, non-gating test (`it.each(PROCESS)('%s requires a #161-adopting REVIEW: CLEARED comment to carry its own session URL (#191)', ...)`) directly after the existing "keeps the limits that did not move" test — it only checks that the sentence is present in prose, the same style as the rest of that describe block.


ugurozsahin:
Closing — opened with a non-compliant branch prefix (`docs/`; the rule is `feature|fix|chore/<n>-<slug>`, and this issue carries no `bug`/`playtest`/`enhancement` label so it is `chore/`). Re-opening immediately from `chore/191-cleared-session-url` with the same commit, same body.

Issue #164, opened 2026-09-16
board: heartbeat


ugurozsahin:
2026-09-17T15:37Z — board: synced 2: #200 status Owner action → In review, #199 status Backlog → In review

Issue #124, opened 2026-09-16
board: heartbeat


ugurozsahin:
2026-09-16T13:02Z — board: synced 2: #114 status In review → In progress, #115 status In review → In progress

ugurozsahin:
Stale duplicate of the `board: heartbeat` pulse, per #125. Last written 2026-09-16T13:02:58Z — the sync switched to writing to issue 164 an hour later (2026-09-16T14:03:13Z) and has been writing there ever since; PR #202 makes that pick deterministic going forward instead of accidental. Closing by hand at the owner's direction, not by the sync itself — CLAUDE.md: "the sync never writes a label, closes an issue or touches a PR".

Session: https://claude.ai/code/session_01UMj9PTws16tSZmfayi68ey

Issue #105, opened 2026-09-16
board: heartbeat


ugurozsahin:
2026-09-14T13:50Z — board: in step — 0 changes, 77 cards

---
_Migrated from #198 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Stale duplicate of the `board: heartbeat` pulse, per #125. Last written 2026-09-16T10:26:37Z (never updated after creation) — the sync has been writing to issue 164 since 2026-09-16T14:03:13Z, and PR #202 makes that pick deterministic (oldest candidate, going forward) rather than accidental. Closing by hand at the owner's direction, not by the sync itself — CLAUDE.md: "the sync never writes a label, closes an issue or touches a PR".

Session: https://claude.ai/code/session_01UMj9PTws16tSZmfayi68ey

PR #198, opened 2026-09-17
Context by layer: CLAUDE.md/ROUTINE-PROMPT.md byte-budget rails (#101)

ugurozsahin:
Part of #101

**Governance PR: tightening.** Adds two new guard rails (enforcement) and moves content to `.claude/rules/*.md` files that already exist — no constraint removed, nothing a run could do before is now permitted with less scrutiny.

## What changed and why

Issue #101's own "Guard rail (tightening)" section asks for two byte-budget rails so the layer-2 migration is a ratchet, not a one-off tidy-up that regrows silently:
- `CLAUDE.md` fails above the byte budget it lands at in this PR.
- `docs/ROUTINE-PROMPT.md` fails if it grows above the size it lands at here.

Neither budget is the issue's eventual target (`CLAUDE.md` is nowhere near ≤4 KB yet — most always-loaded content this issue is about hasn't moved out). That's deliberate: a budget rail records existing debt and only ratchets down as more content genuinely moves to a scoped rule file or a skill.

This PR also moves two bullets out of `CLAUDE.md`'s always-loaded `## Rules` into pointers, so the budget lands below the pre-PR size rather than merely freezing it:
- The `window.__sna` e2e-contract bullet → `.claude/rules/game.md` (`src/game/**`, `src/ui/play.ts`) — that file already carries the real content, landed by PR #179.
- The Android/Capacitor detail inside the "don't add dependencies" bullet → `.claude/rules/android.md` (`android/**`, `capacitor.config.ts`) — also already real content from #179. The general "don't add dependencies without reason" / "no external assets" rule stays in `CLAUDE.md` since it's true everywhere, not path-scoped.

`CLAUDE.md`: 17,644 → 17,609 bytes. `docs/ROUTINE-PROMPT.md`: untouched, budget lands at its current 42,221 bytes as a pure ratchet-lock (stops it growing via some *other* PR while the rest of #101's layer-4 trim is still outstanding).

## What I deliberately did not do

- **Did not touch the three-file-pinned wording** in `CLAUDE.md` (the freeze-history, records-have-readers, #161 adoption and #97 second-item paragraphs) — those are held word-for-word across `CLAUDE.md`/`BACKLOG.md`/`docs/ROUTINE-PROMPT.md` by an existing rail, and none of them are path-scoped content this issue's layer plan asks to move.
[...]

PR #202, opened 2026-09-17
fix: board sync writes to the oldest heartbeat issue, not whichever the API lists first (#125)

### LINKED ISSUE #125: board-sync created a second `board: heartbeat` issue — the pulse two prompts read is now ambiguous
**Found by the dev routine run of 2026-09-14T14:05Z** while doing STEP 1's board-pulse read.

There are now **two open issues titled `board: heartbeat`**, both labelled `watchdog`, and the sync has started writing to the new one:

| issue | created | body | last updated |
| --- | --- | --- | --- |
| #198 | 2026-09-11T17:56:12Z | `2026-09-14T13:50Z — board: in step — 0 changes, 77 cards` | 2026-09-14T13:50:32Z |
| #227 | 2026-09-14T14:05:35Z | `2026-09-14T14:05Z — board: in step — 0 changes, 77 cards` | 2026-09-14T14:05:35Z |

#227 was created with a pulse already in its body at 14:05Z — the same 15-minute tick that would otherwise have rewritten #198. So `scripts/board-sync.mjs` did not find #198 on that run and fell through to its create-a-heartbeat path.

## Why this matters rather than being untidy

Both readers of the pulse look for *the* open issue with that title:

- `docs/ROUTINE-PROMPT.md` STEP 1 — "read the body of the open issue titled `board: heartbeat`".
- `docs/WATCHDOG-PROMPT.md` check 8.

With two, which one a reader picks is now down to API ordering. **#198 will go stale from here** — the sync is writing to #227 — and a reader that lands on #198 sees an ageing timestamp and reports the Mac job dead when it is alive. That is a false "the board sync has stopped" finding aimed at the owner, and it gets worse every hour: by tomorrow #198 reads as a ~24-hour-old pulse.

The inverse is the sharper risk. If the lookup is non-deterministic, then the day the Mac job *really* stops, a reader landing on whichever issue was written last still sees a plausible timestamp for a while — a pulse that can be duplicated is a pulse that can lie in both directions.

## What to look at

`scripts/board-sync.mjs` — its heartbeat lookup and its create path. Two candidates, and the fix differs:

1. **The lookup is fragile.** If it searches by title within a page of open issues (or by `state=open&labels=watchdog` without paging), a busy repository can push #198 off the first page — the repository gained #223–#226 in the preceding hours. Then "not found" is wrong and it creates a duplicate.
2. **It creates unconditionally on any lookup error.** A transient 403/5xx from the Mac should leave the pulse alone until the next tick, not mint a second issue. Creating on failure is the shape that turns one bad request into permanent ambiguity.

Either way the create path wants a guard: **never create a `board: heartbeat` issue when one is already open**, and if more than one is open, write to the oldest and say so in the pulse rather than silently picking.

## Also needs doing by hand

One of the two has to be closed once the script is fixed, and the closure is the owner's or a reviewer's call, not a run's: the routine is told the heartbeat issues are "never work, whatever their labels", and closing a pulse is listed as a finding in its own right. Proposal: **keep #198** (the one both prompts have been reading for three days, and the one with the history) and close #227 with a note pointing at it — but only after the script is writing to #198 again, or the pulse simply goes quiet.

I have not touched either issue.

Note this is not the Mac job being dead: the 14:05Z pulse is fresh and in step, 0 changes, 77 cards. The sync is running. It is the identity of the record that has split.

---

---
_Migrated from #229 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #125

**Not a governance PR** (touches only `scripts/board-sync.mjs`, its `.d.ts`, and its test).

## What changed and why

Three `board: heartbeat` issues are open right now — #105, #124, #164 — and #164 is the only one still being written to. Digging into why: #105 and #124 were created **21 seconds apart** on 2026-09-16, and #164 an hour after #124's last write. `writePulse()`'s old lookup was `open.find((i) => !i.pull_request && i.title === PULSE.title)` — a plain `Array.find` over whatever order the GitHub API happened to return open `watchdog`-labelled issues in. That has two separate problems, and this PR fixes the sharper one:

1. **Which duplicate gets written to is undefined and can change between runs.** `.find` picks array position, not age, so if the API's default ordering ever changes (or already isn't guaranteed stable, which GitHub's docs do not promise), a later run can silently start writing to a *different* stale issue than a reader has bookmarked. #125's own text calls this out as the sharper risk: "the day the Mac job really stops, a reader landing on whichever issue was written last still sees a plausible timestamp for a while — a pulse that can be duplicated is a pulse that can lie in both directions."
2. **A duplicate, once created, is invisible.** Nothing in the pulse line said "there are 3 of these" — a reader has to notice by hand, the way #125 itself was found (by a routine run's STEP 1 read landing on the stale one).

**The fix:** a new pure, exported, unit-testable function `pickHeartbeat(open)` filters to real candidates (title match, never a pull request) and sorts by `created_at` ascending. `writePulse` now always writes to the *oldest* candidate — the one every prompt has already been reading — so a fresh duplicate can never steal the pulse's identity out from under an existing reader. When more than one candidate exists, the pulse line itself now says so: `[N open 'board: heartbeat' issues found (#125) — writing to the oldest, #<n>; the rest are stale duplicates, close them by hand]` — visible to whoever reads the pulse next, instead of silent.

- **Did not close issue 105 or issue 124 from this script, or from this PR.** `CLAUDE.md`: "The sync never writes a label, closes an issue or touches a PR — its only repository write is that pulse." That stays true here; the dupe note tells a *person* to close them, and doing that by hand (with the owner's go-ahead, in the same session as this PR) is a separate action, not part of this diff.
- **Did not add a lock file or other mechanism to make issue creation atomic.** GitHub's REST API has no compare-and-swap for issue creation, so a *guaranteed* fix would need an external lock (e.g. a file on the Mac, or a repository-side marker) — genuinely harder, and I could not reconstruct with confidence exactly how #105/#124 raced 21 seconds apart (two overlapping invocations of the script, one scheduled and one manual, is the likeliest read of the timeline, but I have no log access to confirm it). This PR fixes the *consequence* that matters — identity drift and silent duplication — deterministically and testably; eliminating every possible race that could ever create a fourth duplicate is a separate, harder problem I'm not guessing my way through here.
- **Did not touch `writePulse`'s create-vs-patch branching, `pulseNeeded`, or anything about *when* a pulse is due.** Only *which* issue it targets changed.

## Tests run

- `npx tsc --noEmit` — clean (added `HeartbeatIssue` to `scripts/board-sync.d.ts`, matching the existing hand-maintained-declarations pattern the file already uses for `plan`/`pulseNeeded`/etc.)
- `npm test` — 1082/1082 (30 files), including 2 new cases in `tests/unit/board-sync.test.ts` (49 total there, up from 47): one reproducing #125's exact #105/#124/#164 timeline in three different input orderings and asserting the oldest always wins; one confirming a pull request carrying the same label is ignored and an empty/no-match input returns `[]`.
- `npm run build` — clean, `sna-848d36323323`
- e2e correctly skipped — diff touches only `scripts/board-sync.mjs`, `scripts/board-sync.d.ts`, `tests/unit/board-sync.test.ts`; none of #96's game-reaching paths.
- Proved red before green, twice: reverted `pickHeartbeat` to the old unsorted `.filter` (no `.sort`) and confirmed the new "picks the oldest" test failed with exactly the wrong order (`[164, 124, 105]` instead of `[105, 124, 164]`); restored, reran — 49/49 green, file byte-identical to before the mutation (`diff` confirmed).

## Guard-rail budgets

No budget rail moved. Two new, independent test cases added to a new `describe` block in `tests/unit/board-sync.test.ts`, directly testing the new pure function — no mocking of `fetch` needed, matching this file's existing style of keeping the network shell thin and the logic pure/testable.


ugurozsahin:
Reviewed `0d87638` (merges cleanly with current main `b227424`, no conflict). `npm test` 1088/1088, `npx tsc --noEmit` clean, `npm run build` clean. e2e correctly skipped (diff touches only `scripts/board-sync.mjs`/`.d.ts` and `tests/unit/board-sync.test.ts` — none of #96's game-reaching paths). Confirmed this doesn't collide with #198's new CLAUDE.md/ROUTINE-PROMPT.md byte budgets (it touches neither file).

Ran the three vendored review agents (#99):
- **`pr-test-analyzer`**: `pickHeartbeat`'s ordering-independence, PR-exclusion and empty-input cases are well covered. Gap (non-blocking): `writePulse`'s own `dupeNote` string (the `[N open ... issues found ...]` text actually appended to the pulse body) isn't exported or tested anywhere — only the pure `pickHeartbeat` sort/filter is. Also no explicit tie-break case for identical `created_at` timestamps.
- **`silent-failure-hunter`**: one real but low-likelihood risk — `new Date(a.created_at) - new Date(b.created_at)` would silently degrade to engine-dependent ordering on a malformed/missing `created_at` (`NaN` from the subtraction), with no validation or log. GitHub's REST payload always has this field, so not blocking, but worth a follow-up guard.
- **`type-design-analyzer`**: minor nit — `pull_request?: object` should be `pull_request?: unknown`, since only presence (not shape) is ever checked. Not blocking.

None of these are correctness bugs in what ships — they're follow-up hardening (untested dupeNote string, a defensive NaN guard, a type nit). Filing a small `debt`-labelled follow-up issue rather than blocking a correct, well-tested fix on them. Squash-merging.

Session: https://claude.ai/code/session_01C2pKTDxG5Ajirpyv3FqQdZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2pKTDxG5Ajirpyv3FqQdZ)_

---

PR #208, opened 2026-09-17
Fix a stalled REVIEW: CHANGES REQUESTED block before new work; stop gh's duplicated footer (#204, #207)

ugurozsahin:
**Tightening.** Adds two process rules a run must follow; nothing a run was permitted to do before is
prevented, and no existing rail is weakened.

Two related findings from today's session, both about how the review/fix cycle actually behaves on this
repo right now.

**#204 — nothing ever pushed a fix for a block it did not set.** STEP 2's own filter correctly treats a
`REVIEW: CHANGES REQUESTED` PR a run cannot clear as "not waiting for a review this run could do" and moves
the run on to STEP 3 — that part is working as designed. But nothing then assigned anyone to push the fix
either: "the author pushes a fix" assumes continuity a stateless, anonymous routine run does not have.
Evidence from today: PR #197's block sat from 14:52Z with its fix already pushed at 15:12Z and still
uncleared past 17:00Z; PR #201 sat blocked from 16:45Z with no fix pushed at all; a run opened an unrelated
pull request for issue 101 in between — correct per STEP 2 — but nobody was also moving the two open blocks
forward.

**#207 — a `gh`-posted comment or body can carry a duplicated, unrelated footer.** Every comment posted via
`gh issue comment`/`gh pr comment`/`gh pr create --body` in a cloud run comes back with an `---` rule and
`_Generated by [Claude Code](...)_` appended — sometimes twice, with two different links — on top of
whatever the comment itself says, including this repo's own `Session:` line. Every comment posted with the
REST API directly (this PR's own issues, and #201/#202/#203) carries none of it.

**Where the fix lands, and why this version is different from the first push.** Both rules now live in
`docs/ROUTINE-PROMPT.md` only — the one file every run demonstrably reads in full every run (STEP 1) — not
`BACKLOG.md`. The first version of this PR put them in `BACKLOG.md` instead, reasoning only from "which file
has no byte budget"; that was wrong — the owner is retiring `BACKLOG.md`, so it is not a stable home for a
behavioural rule, and nothing established that a run reads it as reliably as `docs/ROUTINE-PROMPT.md` going
forward anyway. `docs/ROUTINE-PROMPT.md` was pinned at zero headroom by #101's byte-budget rail (PR #198,
landed today), so landing the two rules there meant first trimming an equal or greater number of bytes
elsewhere in the same file — cut from historical incident narration (which private-archive PR, what time,
[...]

PR #203, opened 2026-09-17
Context by layer: move the Guard rails paragraph out of CLAUDE.md (#101)

ugurozsahin:
Part of #101

**Governance PR: tightening** (moves content out of the always-loaded `CLAUDE.md` into a scoped `.claude/rules/*.md` pointer that already exists — a net reduction, and the budget rail's own ceiling moves down with it; nothing a run is permitted to do changes).

## What changed and why

The next safe slice of #101's Layer 4 trim: `CLAUDE.md`'s "Guard rails" bullet duplicated content that `.claude/rules/guardrails.md` (created by an earlier #101 slice) should own but didn't yet carry — the specific mistake list (render frame rate, screen teardown, CSS collisions, `as any`, comparator shuffles, `shadowBlur`, dependency allowlist, listener pairing, British English) and the "text/DOM checks, not proofs" caveat.

- `CLAUDE.md`'s bullet is now a one-line pointer to `.claude/rules/guardrails.md` (`tests/unit/guardrails.test.ts`, `scripts/**`), the same shape as the `game.md`/`android.md`/`curriculum.md` pointers already there.
- `.claude/rules/guardrails.md` gains the mistake list and the "text/DOM checks, not proofs" sentence it was missing — nothing is dropped, only moved.
- `CLAUDE.md`: 17,609 → 17,175 bytes. The #198 budget rail's `CLAUDE_MD_BUDGET` constant moves down to match (17,609 → 17,175) — a genuine ratchet, not a raise.

## What I deliberately did not do

- **Layer 0 (the three `PreToolUse` hooks)** — not attempted, same reason #178/#198 gave: a routine/cloud session cannot reload the settings watcher or open `/hooks` to prove a hook fires live, and shipping an unverified one risks silently breaking tool permissions for every future session.
- **`docs/ROUTINE-PROMPT.md`'s own copy of the mistake list** (STEP 2's "Guard rails." paragraph) — left untouched. It's a separate flow document, not `CLAUDE.md`'s always-loaded context, and folding it into this slice would grow the diff past one clean, reviewable change. A follow-up slice can point it at the same rule file if that's judged worthwhile.
- **The other three `CLAUDE.md` bullets #101 still lists as outstanding** (the freeze paragraph, the governance/review-block paragraphs) — each needs its own careful read against the three-file-pinned-wording rail (`#161`/`#97`/`#98`/freeze-history) before any of it can move; not rushed into this PR.
[...]

PR #201, opened 2026-09-17
Session URL and a content floor for every comment and issue, not just REVIEW: markers (#199, #200)

ugurozsahin:
Closes #199
Closes #200

**Governance PR: tightening.** Adds two sentences to the shared #161-adjacent paragraph (`CLAUDE.md`, `BACKLOG.md`, `docs/ROUTINE-PROMPT.md`), extends `review-pr/SKILL.md` and `open-pr/SKILL.md` to describe the shapes this already asked of the two `REVIEW:` markers, and adds two new guard-rail tests. It does not touch `scripts/review-gate.mjs` — `review-gate`'s gating output is exactly the same before and after this PR.

## What changed and why

#199 and #200 came out of auditing every open PR for the gap #189/#190/#191/#193 had already partly closed: a `REVIEW: CHANGES REQUESTED`/`REVIEW: CLEARED` comment carries its own session URL, but nothing else a session writes here did. Concretely, today: PR #178's and PR #187's own fix-push comments shipped with zero session identification and had to be patched by hand once found; no issue body checked (#67, #189, #191, #194, #195) carried one either; PR #197's fix-push comment only had one because it happened to pick up the CLI's own auto-footer, which a plain API call does not add.

**#199** makes the rule universal instead of special-cased to two comment types: every comment a session posts to any issue or pull request, every issue body, and every pull request body, ends with its own `Session: https://claude.ai/code/session_<id>` line — the CLI's auto-footer is explicitly called out as not a substitute (it names whoever's tool call produced the post, not necessarily who wrote the words above it, and a raw API call gets no footer at all).

**#200** writes down the content shape that fix-push comments and issue bodies already tended to have organically (seen across #178/#187/#197's fix-push comments and #194/#195/#199/#200's own bodies): a fix-push comment opens `Pushed <sha>, addressing <what>`, resolves each blocking finding by the review's own numbering, states tests run, and closes `Ready for re-review` (never `REVIEW: CLEARED` — that stays the reviewer's own mark); an issue proposing a fix states the problem with evidence, a `## Proposed fix` section, and a `## What this deliberately does not do` section — the same two-sided shape `open-pr/SKILL.md` already asks of a PR body. Everything else (status notes, the "starting on this" announcement) only has to clear the floor: say the actual point up front, sign with `Session:`.

[...]

PR #197, opened 2026-09-17
Onboarding wizard: split choose-ninja and name onto separate screens (#67)

ugurozsahin:
Closes #67

**Not a governance PR** (touches only game UI/tests/scripts).

## What changed and why

#67's acceptance criteria call for the first-run wizard's ninja pick and name entry to be two separate
screens, each fitting one viewport without scrolling — the part PR #188 (the progress-rail slice) explicitly
deferred as "real work that deserves its own reviewable PR".

`src/ui/avatar.ts`'s combined `avatarScreen` is now two functions:
- `chooseNinjaScreen` (step 1 of 3) — the eleven-card grid, gated by a new `#next` "Continue ➡️" button that
  needs only an avatar.
- `nameScreen` (step 2 of 3) — the chosen ninja shown alongside the name field, keeping the existing `#go`
  button and `canStart(avatar, name)` gating.

`introScreen` (from #181) is now step 3 of 3. `changeAvatarScreen` (returning players via `#change-av`) is
unchanged — ninja-only, no name field, no progress rail.

`main.ts` gains an `onboard-name` history state between the existing `avatar` and `onboard-intro` ones,
wired into `nav`/`popstate` the same way `onboard-intro` already was. Finishing the wizard now unwinds
**two** pushed history entries (`history.go(-2)`) instead of one — the wizard is one entry deeper than before,
and `history.back()`'s old single pop would land back on the name step instead of the map. Caught by every
`…and lands on .home` e2e test failing identically once I ran the suite, before this ever reached a PR.

## #110, strengthened rather than just moved

The original #110 rail pinned that `.name-row` renders *before* `.avatar-grid` in one combined screen's
markup. Now that they're separate screens, the rail checks that `chooseNinjaScreen`'s body never contains
`.name-row` and `nameScreen`'s body never contains `.avatar-grid` — the name field has nothing left to be
buried under, on any viewport, rather than merely being ordered above it on one long page.

## Deliberately not done here

- No wording/copy changes to the introduction (#181's) — unchanged.
- No new artwork — the name step reuses the existing avatar art (`.figure img`) at a smaller size via a new
  `.name-preview` class, following `design-language`'s "reuse a class before inventing one".
- Back/forward, first-run-only gating, choices-saved-as-made, read-aloud and keyboard/focus accessibility
  were already correct from #181/#188 and are unchanged in kind — this PR extends them to a third
  history state rather than reworking them.

## Look change — owner-approval

[...]

PR #210, opened 2026-09-17
fix: review-gate flags a #161-adopting REVIEW: CLEARED with no session URL of its own (#195)

### LINKED ISSUE #195: review-gate should also flag a #161-adopting REVIEW: CLEARED comment with no session URL
Follow-up from PR #193's review (silent-failure-hunter finding, not blocking that PR — see the review comment on #193).

#189/#190 made `scripts/review-gate.mjs`'s `blockState()` flag a *blocking* `REVIEW: CHANGES REQUESTED` comment that carries no session URL (`'no session URL — unadoptable per #161'`). #191/#193 documented the matching rule for the *clearing* side — a `REVIEW: CLEARED` comment that adopts a stale block under #161 must carry the clearing session's own URL too — but deliberately left `blockState()` unchanged, since PR #171's two existing `REVIEW: CLEARED` comments are both unmarked and a retroactive gating check would incorrectly re-flag an already-legitimately-cleared PR.

So today: a future `REVIEW: CLEARED` comment that adopts a stale block and omits its own session URL passes CI silently — nothing turns red, and the gap #189 just fixed on the blocking side is still open on the clearing side.

## Proposed fix
Extend `blockState()` (or a sibling check) to detect when the newest `REVIEW: CLEARED` comment is itself a #161 adoption (i.e. its body identifies another session's block as being adopted, per the #161 language) and add a loud reason when that comment carries no session URL of its own — mirroring #189's shape, but for the clearing side. Needs care to stay non-retroactive for already-cleared PRs the way #193's version was, and to distinguish a #161-adoption clear from an ordinary clear by the original reviewer (which does not need to name a session at all).

Not urgent — this is a documentation gap with no live incident yet (only #171's already-settled comments), unlike #189/#160/#179 which were live and current.

ugurozsahin:
## Problem

`#161` lets a new run adopt (clear) a stale block another reviewer set, but only if the adopting `REVIEW: CLEARED` comment names the conditions **and carries its own session URL** — `docs/ROUTINE-PROMPT.md` and `CLAUDE.md` both say so in the exact phrase: "When that REVIEW: CLEARED comment is the adoption itself, it carries the clearing session's own URL too." (`#191`)

`#189` made the *blocking* side of this enforceable: `scripts/review-gate.mjs`'s `blockState()` already flags an open `REVIEW: CHANGES REQUESTED` with no session URL as "unadoptable per #161." But nothing ever checked the *clearing* side — an adopting `REVIEW: CLEARED` with no session URL of its own currently clears the block silently and passes review-gate with no reason logged at all. `#191`'s rule has lived only in prose since it was written.

Evidence: PR #171's two historical adoption-clear comments (both open with "Clearing another reviewer's block, adopted under #161.") predate `#191` and never carried a session URL — exactly the gap `#195` describes. That PR is merged and closed, so this fix is non-retroactive there; and none of the 4 PRs open right now (#208, #203, #201, #197) has any `REVIEW: CLEARED` comment at all, so the new check can't retroactively flag a live PR either.

## Proposed fix

- New export `isAdoptionClear(body)`: true when a comment is a `REVIEW: CLEARED` (`isCleared`) whose text also matches `/another reviewer'?s block/i` — the exact phrase `#161` itself uses, and the one every real adoption comment on record opens with. An ordinary clear by the reviewer who set the block never needs to say this, so the phrase's presence is a safe, non-accidental signal.
- `blockState()` now also tracks the newest clear's own body (`clearedBody`), and adds a new reason — `"REVIEW: CLEARED adopts another reviewer's block with no session URL of its own — unmarked per #161/#191/#195"` — only when: the clear is the PR's *current* state (not superseded by a later open block, same "newest marker wins" rule `#189` already established), and it is an adoption clear, and it has no session URL of its own.
- `scripts/review-gate.d.ts` gains the `isAdoptionClear` declaration.
- Ordinary clears (no adoption language) are never flagged, regardless of session URL — the rule is specific to `#161` adoption, not clears in general.

## Tests

`tests/unit/review-gate.test.ts` gains a new `describe` block mirroring the existing `#189` session-URL block: `isAdoptionClear` on real (PR #171 phrasing) and synthetic bodies; the new reason firing on an adoption clear with no URL; the reason absent when the adoption clear does carry a URL; the reason absent for an ordinary clear; and the reason absent when the adoption clear is superseded by a later block (non-retroactive, same as `#189`).

Verified red-then-green: reverted `scripts/review-gate.mjs` to `origin/main` and confirmed the 6 new/affected assertions fail, then restored and confirmed all pass. Full suite: 1098/1098 passing. `tsc --noEmit` clean. `npm run build` clean.

- Does not retroactively flag any closed or currently-open PR (verified above).
- Does not change `#189`'s existing blocking-side check or its reason text.
- Does not touch the `#161` adoption conditions themselves (age, silence window, re-derivation) — only adds the missing enforcement for the session-URL requirement `#191` already wrote down.

Closes #195.


ugurozsahin:
Merged as `99f2fcc` (squash), closing #195.

Verified on head `17d5e24`: checked out the branch, `npx tsc --noEmit` clean, `npm test` 1098/1098 (matches PR claim), `npm run build` clean. Re-merged onto the current `main` tip after this session's own two earlier merges (#203, #208) — clean, no conflict, full suite still 1102/1102 with build clean on that merged tree. Diff touches only `scripts/review-gate.mjs`/`.d.ts` and `tests/unit/review-gate.test.ts` — no game-reaching paths, e2e correctly out of scope.

**Attacked the fix directly**: reverted `clearNeedsSession` to its pre-fix value (`false`), confirmed exactly the one new test this PR adds (`adds a distinct reason when the current clear is an adoption with no session URL of its own`) goes red while the other 9 new cases stay green — the rail catches precisely the regression it names, nothing more, nothing less. Restored the file, confirmed byte-identical via `diff`, reran green.

Ran the three vendored review agents (#99), all converging on the same class of finding — untested **edge cases that lean toward over-blocking, never under-blocking**:
- `pr-test-analyzer`: `isAdoptionClear`'s regex isn't anchored to the adoption context, so a `REVIEW: CLEARED` comment that mentions "another reviewer's block" in an unrelated/quoting sentence would false-positive as needing a session URL it doesn't need. Also untested: an original block that itself lacked a session URL (#189) combined with an adoption clear that also lacks one, and a same-timestamp clear/request tie.
- `silent-failure-hunter`: confirmed `clearNeedsSession` is unconditionally pushed into `reasons` (no missing-push path) and `clearedBody` is always set alongside `cleared` in the same loop iteration (no null-deref path). Same regex-false-positive gap independently: a CLEARED comment coincidentally matching the adoption phrase with **no prior REQUESTED at all** would still be flagged.
- `type-design-analyzer`: `cleared`/`clearedBody` are parallel nullables that must be kept in sync by hand (same pattern the existing `requested`/`requestedHasSession` pair already uses) — a `{ at, body } | null` shape would let the type system enforce that instead. Not a defect, a nice-to-have.

None of these are blocking: the shared finding is a narrow false-positive (an ordinary clear wrongly asked for a session URL it doesn't strictly need), never a false negative that would let a real unattributed adoption pass silently — and #199/#200 already mandate a session URL on every comment regardless, so the practical exposure is low. Reachability check per the review skill: the false-positive path requires a `REVIEW: CLEARED` comment that isn't an adoption but happens to contain the literal phrase "another reviewer's block" — plausible in the abstract, not something any real comment on this repo has ever done. Worth tightening in a follow-up if it ever bites; not filing an issue for a hypothetical that hasn't occurred.

Governance/three-file check: this PR touches only `scripts/review-gate.mjs`/`.d.ts` and its test file, none of the three governance-pinned files, so the three-file rule doesn't apply here.

CI green on this head (`review-gate`: "No open review block"; `CI` run 35260869414 success after undrafting re-ran it).

Session: https://claude.ai/code/session_01TDtBF6hcbz5EzTGq2zB2NT

---

Issue #98, opened 2026-09-16
Stop writing WORKLOG.md: archive it, move the checks into the heartbeat issue


ugurozsahin:
Owner (session, 2026-09-10): **stop writing `WORKLOG.md`.** Archive what exists and move the only part that has no other home into the run's heartbeat issue.

## The measurement behind it
- `WORKLOG.md` is **51 KB today**, up from 1.5 KB on 2026-09-07 — roughly **17 KB a day, ~3 KB per run**.
- The previous file reached **94.8 KB and killed a run on the token limit**; it was rotated by hand to `docs/worklog/2026-09.md` at 2026-09-07 17:29 (`8c0a255`) and the counter restarted. At the current rate the same line is about **three days away**.
- **There is no rotation logic anywhere.** No script, no workflow. The one rotation that has ever happened was manual.
- **Nobody reads it, by rule.** `CLAUDE.md`: *"Nothing reads WORKLOG.md — it is a write-only record for humans."* `docs/ROUTINE-PROMPT.md` STEP 1: *"**Do not read WORKLOG.md.**"* Twelve-plus writers a day, zero readers.

Most of what it holds is a narrative copy of things recorded better elsewhere: what changed and why is the pull request and its review; when is the git history; what is in flight is the open PR list.

## What has to be preserved
One part genuinely has no other home: the run's **checks** — was last night's CI on `main` green, how old is the watchdog pulse, was anything anomalous and deliberately not acted on. That is operational telemetry, and losing it silently would be worse than the file.

Move it to the run's **`routine: heartbeat` issue**, and note the trap: that issue's body must be **overwritten each run**, never appended. A fixed-size snapshot — last run time, each check and its verdict, anything the run decided not to do and why. Appending there would recreate exactly the problem this issue exists to end, in a place that *is* read.

## What to change
- `docs/ROUTINE-PROMPT.md` STEP 5 (RECORD) stops writing the file and writes the heartbeat snapshot instead; STEP 1's "do not read it" instruction goes with it.
- `CLAUDE.md` and `BACKLOG.md` lose the WORKLOG sentences (three-file rule — same wording, one PR).
- `docs/WATCHDOG-PROMPT.md` — same treatment for the watchdog's own record.
- `scripts/seed-issues.py` mentions it; check what for.
- **`WORKLOG.md` moves to `docs/worklog/` alongside the earlier archive.** It is not deleted: it is the record of the project's first week and reads as history. Nothing appends to it again.
[...]

PR #217, opened 2026-09-17
Context by layer (#101): Layer 0 - three PreToolUse hooks, no context cost

ugurozsahin:
Part of #101

**Not a governance PR in the tightening/loosening sense that matters** — it adds a check
(`.claude/settings.json`'s three hooks plus the new guard rail pinning them), so under CLAUDE.md's
rule it is ordinary tightening: enforcement added, no constraint removed, no gate weakened.

## What changed and why

#101 asks for three Layer 0 (PreToolUse hook) enforcements alongside the seven `.claude/rules/*.md`
files already landed by earlier slices of this issue: absolute rules that should cost no context
on every turn, because CLAUDE.md currently only spends a sentence on each and that sentence has
still been broken. This PR is that slice: `.claude/settings.json`, committed so a cloud/routine
session gets it the same way it gets committed rules and skills.

- **git push** — denies `--force`/`-f`/`--force-with-lease`, gated by the hook's own
  `if: "Bash(git push *)"` so it only runs on an actual push.
- **The owner's verdict markers** — denies a `gh` or `curl`-to-`api.github.com` command whose text
  carries `OWNER: APPROVED` or `OWNER: REJECTED`.
- **WORKLOG.md (#98)** — denies a Bash `>>` append and a Write/Edit, both scoped to the file at the
  repo root only; `docs/worklog/*.md` is unaffected.

Each hook's `permissionDecisionReason` names which rule it enforces and where it is written, per
the issue's own requirement.

## A real false positive, found and fixed before this landed

The marker hook's first draft matched any Bash command containing either marker string, with no
`if:` filter. Pipe-testing it against synthesized stdin caught the intended cases, but the very
next thing it did — for real, live, in this session — was deny **this commit**, because the
commit message itself discusses the markers in prose (as this PR body also does). A blanket
substring match cannot tell a forgery attempt from a sentence *about* the mechanism, and this
repository's own PR/commit style talks about `OWNER:`/`REVIEW:` markers constantly.

Fixed by scoping the hook to the two shapes that can actually post a comment to GitHub over Bash:
`if: "Bash(gh *)"` and `if: "Bash(curl *)"` (the latter additionally checks for
`api.github.com`), rather than every Bash command. `git commit`, `git log`, editing a doc, or
writing a PR body that quotes the markers are all outside the `if:` filter entirely, so none of
them reach the hook's own grep. Re-tested both scoped hooks: still deny a `gh`/`curl` command that
actually carries a marker, still allow one that doesn't.

## Known gap — left as a follow-up

This only covers the Bash/CLI path (`gh`, `curl`). A session posting a comment via an MCP GitHub
tool call directly (`add_issue_comment`, `pull_request_review_write`, `issue_write`, …) is not a
Bash command at all, so a `matcher: "Bash"` hook cannot see it — and this session posts every
comment that way, not via `gh`. Covering that needs a separate `PreToolUse` hook per MCP tool name,
checking that tool's own body/comment argument, which is a larger, separate change than this
slice. Filing as a fast-follow rather than attempting it incompletely here.

- **Did not also enforce the branch-name half of the git-push item** (`main` or
  `feature|fix|chore/<n>-<slug>`). `ci.yml`'s existing `branch-name` job already checks this
  server-side on every PR; a client-side re-implementation risks a false positive blocking a
  legitimate push over a parsing edge case, for a check that already exists downstream. Left for a
  follow-up if the owner wants belt-and-suspenders here.
- **Did not build the issue's "Verification" section** — an `InstructionsLoaded` hook demonstrating
  a docs-only run loads only `CLAUDE.md`/`governance.md` and a curriculum run loads only
  `curriculum.md`. That needs two actual separate session runs to produce logs from, not something
  one session can manufacture for itself; left for a follow-up slice, same as PR #178 deferred
  these very hooks in the first place.
- **Did not touch layer 4** (trimming `docs/ROUTINE-PROMPT.md`'s own remaining freeze/governance
  prose into pointers) — separate, larger slice.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1116/1116 (30 files, 4 new) |
| `npm run build` | clean |
| e2e | correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts` |

Each hook command was pipe-tested directly with synthesized stdin JSON before being committed
(the `update-config` skill's own verification method): force flag denied, `force-with-lease`
denied, a plain push allowed, a branch name containing the word "force" not mistaken for the flag;
both markers denied via `gh` and via `curl` to `api.github.com`, an unrelated `gh`/`curl` comment
allowed; `WORKLOG.md` append/write denied, `docs/worklog/*.md` append/write allowed.

## Guard rail (tightening)

A new `describe` block in `tests/unit/guardrails.test.ts` fails if `.claude/settings.json` is
missing, or if any of the three hooks isn't declared — checked by the distinguishing text each
hook's command must contain (the `if:` filter for the push hook, both marker strings for the
verdict hook, `WORKLOG.md` plus a deny decision for the third), not by re-implementing the
matching logic itself. Proved red first: moved the file aside, watched all four new assertions
fail with a clear `ENOENT`, restored it, watched all four pass again.

`node scripts/review-gate.mjs body.md`: `closes: nothing`.

Session: https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS

https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS

---
_Generated by [Claude Code](https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS)_


ugurozsahin:
Pushed `cbb1e56`, correcting my own mistake: the first commit's message described scoping the owner-marker hook to `gh`/`curl` via `if:` (the false positive fix in the PR body above), but the `if:` filter was added to the working tree *after* I'd already staged the file, so the commit itself still had the broad, unscoped hook — my commit message didn't match what I'd actually committed.

This push adds the missing `if:` filters (`Bash(gh *)`, `Bash(curl *)`) to both marker-hook entries, matching what was already tested and described. Verified again on this exact tree: `npx tsc --noEmit` clean, `npm test` 1116/1116, `npm run build` clean.

Session: https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS

---

ugurozsahin:
Note on the `review-gate` commit-status failure on this PR: it's expected, not a bug to fix with a push. This repo's `review-gate` check is designed to read `failure` while a PR is a draft (`CLAUDE.md` / `docs/ROUTINE-PROMPT.md` STEP 2 rule 3) — the Actions workflow run itself is green regardless, by design, so the failing commit status is the gate working correctly. This PR is intentionally a draft: I opened and developed it, so the developer/reviewer split means a different session reviews and merges it, not me.

Actual CI on the current head (`cbb1e56`) is green: `test`, `branch-name` and `gate` (the Actions workflow job, not the `review-gate` status) all completed successfully. No third-party check suite is failing or pending. Nothing here needs a fix — it clears once a different session reviews and undrafts it.

Session: https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS

---

ugurozsahin (commented):
placeholder

---

ugurozsahin:
Saw the review submitted just now (state: Commented) — its body is just "placeholder", with no line comments attached, so there's nothing concrete here for me to act on yet. Not treating it as a block: it isn't a draft-mark + `REVIEW: CHANGES REQUESTED` (this repo's only blocking mechanism, per `CLAUDE.md`), so this PR stays exactly as it was — draft, CI green on `cbb1e56`, awaiting a different reviewer's actual pass.

If real feedback was meant to go with this, happy to address it once it's posted — flagging now rather than silently ignoring the review event.

Session: https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS

---

ugurozsahin:
REVIEW: CHANGES REQUESTED

(The "placeholder" GitHub review just above this comment was a tool-call mistake on my part — I called `pull_request_review_write` with unverified parameters and it submitted before I'd written real content. Please disregard it entirely; this comment is the actual review, and per this repo's protocol a review blocks via a draft + a `REVIEW:`-prefixed comment, never a formal GitHub review object anyway.)

CI is green on the current head (`cbb1e56`: `test` and `branch-name` both succeed; e2e correctly out of scope per #96 since the diff only touches `.claude/settings.json` and `tests/unit/guardrails.test.ts`). `npx tsc --noEmit`, `npm test` (1116/1116) and `npm run build` all independently reproduce clean on this branch. The PR is already correctly a draft; `review-gate` reading red is that mechanism working as intended, not a finding.

The problem is that two of the three Layer-0 hooks this PR exists to add have real, concrete bypasses of the exact thing they claim to prevent — found by `silent-failure-hunter` and independently verified by me by tracing the regex against the literal strings below:

1. **Force-push denial is bypassed by a combined short flag.** `git push -uf origin main` (an entirely ordinary git idiom for "set upstream and force") is **allowed**. The pattern `(^| )-f( |$)` requires `-f` as its own space-bounded token; the substring `-f` never occurs standalone in `-uf` (it's `-u` then `f`), and in `-fu` it occurs but isn't followed by a space/end-of-string. Given force-push is one of this project's explicit "never" rules, a hook meant to backstop that which a one-character-reordered flag defeats isn't backstopping it.
2. **The WORKLOG.md Bash-append hook is bypassed by an absolute path or a quoted filename** — `echo x >> /home/user/sky-academy/WORKLOG.md` and `echo x >> "WORKLOG.md"` both slip past `>>[[:space:]]*(\./)?WORKLOG\.md`, which only allows no prefix or exactly `./` before the filename. Since this environment's own tool conventions push absolute paths, and quoting a filename in a shell append is completely ordinary, these aren't exotic inputs.

`pr-test-analyzer` separately found that the new guard rail's own test coverage wouldn't catch a regression in either direction: deleting either the `gh`-scoped or the `curl`-scoped OWNER-marker hook leaves the rail green (a single `.find()` checks that *some* command contains both marker strings, not that both hooks independently exist); the `api.github.com` scoping on the curl hook and the `if:` gates on both marker hooks are never asserted at all; and nothing in the rail actually executes any hook's shell script against synthesized stdin — every assertion is JSON-shape/substring only, so a broken `jq` path or inverted `grep` logic in any of the four hook bodies would leave all four new tests green (this is also how #2 above ships undetected by the new rail, since the rail only checks the regex text contains `WORKLOG.md`, never what it does or doesn't match).

`type-design-analyzer` found nothing blocking — the `any`-typed test helpers are consistent with this file's existing house style and every risky access is already backed by an explicit, descriptively-named assertion; a loose `interface` for the settings shape would be a nice-to-have, not a gap.

What needs to change before this is mergeable:
- Fix the force-push regex to catch a `-f` bundled into a combined short-option cluster (`-uf`, `-fu`, etc.), not just a standalone token.
- Fix the WORKLOG.md Bash-append regex to catch an absolute path to the same root file and a quoted filename.
- Strengthen the guard rail to actually pin what regressed: independently assert both the `gh`- and `curl`-scoped marker hooks exist (and their `if:`/`api.github.com` gates), and add cases to the WORKLOG.md test proving the fixed regex denies the absolute-path and quoted-filename forms above (and still allows `docs/worklog/*.md`).
- Given three of these findings only surfaced from actually tracing the regex/logic rather than checking the JSON shape, consider whether the rail should exercise the hook shell scripts directly (piping synthesized stdin, as the PR body describes doing manually) rather than relying solely on text/shape assertions — that's what would have caught 1 and 2 mechanically rather than by review.

Not blocking, left to your judgement: the Write/Edit hook's `[ "$dir" = "$(pwd)" ]` string comparison (not path-canonicalized) that `silent-failure-hunter` flagged as a possible false-negative under a git-worktree checkout — lower confidence, and this repository's routine sessions don't currently use worktrees for this file, but worth a look if #101's later layers ever do.

---

ugurozsahin:
Pushed `63be8e2`, addressing both blocking findings and the guard-rail weakness from the review above.

**1. Force-push bypass (`-uf`/`-fu`).** Fixed: `(^| )-[a-zA-Z]*f[a-zA-Z]*( |$)` now catches any single-dash cluster containing `f`, not just a standalone `-f` token. Safe for `git push` specifically — none of its other short options (`-v`, `-q`, `-u`, `-d`, `-n`, `-o`) contain the letter f, so this can't false-positive on a legitimate combination.

**2. WORKLOG.md bypass (absolute path / quoted filename).** Fixed: the command's quotes are stripped first (`tr -d '\047"'`), then matched against `>>[[:space:]]*([^[:space:]]*/)?WORKLOG\.md(...)`, so any path prefix ending in that basename is caught. `docs/worklog/*.md` still isn't, since its basename never matches.

**3. Guard rail didn't independently pin the gh/curl marker hooks, and never executed a hook's actual logic.** Both fixed: the OWNER-marker structural check is now two independent assertions (one per hook, each also checking its `if:` gate and the curl hook's `api.github.com` scoping), and a new `describe` block runs every hook's real shell command via `execFileSync('bash', ['-c', command])` against synthesized stdin, asserting on actual stdout. Proved these new execution tests catch both bypasses: stashed the fix, ran only the new block, watched the `-uf` and absolute-path cases fail with the exact review-reported bypass, restored the fix, watched them pass.

**Not blocking, not addressed** (per the review's own note): the Write/Edit hook's `[ "$dir" = "$(pwd)" ]` comparison isn't path-canonicalized — left for a future look if #101's later layers use worktrees for this file.

Tests run on this exact tree: `npx tsc --noEmit` clean, `npm test` 1122/1122 (30 files, 10 in the two layer-0-hooks blocks, up from 4), `npm run build` clean. Re-verified every previously-passing case alongside the two fixes — no regressions.

Ready for re-review.

Session: https://claude.ai/code/session_01GFdEyEpV4NZ39SvmUvL7iS

---

ugurozsahin:
REVIEW: CLEARED

Independently re-verified both fixes on `63be8e2`, not just read the diff:

1. **Force-push bypass** — ran the actual hook command (`bash <hook> <<< '{"tool_input":{"command":"git push -uf origin main"}}'`) myself: denies `-uf` and `-fu`, still allows a plain `git push -u origin fix/1-force-push-test` (branch name containing the word "force" isn't mistaken for the flag).
2. **WORKLOG.md bypass** — same direct execution: denies `>> /home/user/sky-academy/WORKLOG.md` and `>> "WORKLOG.md"`, still allows `>> docs/worklog/2026-09.md`.
3. Guard rail: confirmed the two OWNER-marker hooks are now asserted independently (with their `if:`/`api.github.com` gates), and the new execution-based `describe` block genuinely runs each hook's real shell command via `execFileSync` rather than shape-checking.

Also reproduced clean myself on this exact tree: `npx tsc --noEmit`, `npm test` (1122/1122), `npm run build`. CI on `63be8e2` is green (`test`, `branch-name`, `gate`).

Marking ready for review, which will re-run CI per #88 — merging once that fresh run is green.

---

ugurozsahin:
Merged as `3114485` (squash). The fresh CI run triggered by marking this ready for review (run 35285308545, head `63be8e2`) was green — `test` and `branch-name` both succeeded, `review-gate` read "No open review block" — so this wasn't merged on a stale tick.

Final state: `npx tsc --noEmit` clean, `npm test` 1122/1122, `npm run build` clean, e2e correctly out of scope (#96). Both bugs from the `REVIEW: CHANGES REQUESTED` are fixed and independently re-verified by me against the real hook commands, not just re-read: `git push -uf`/`-fu` now denied, an absolute-path or quoted `WORKLOG.md` append now denied, `docs/worklog/*.md` and a plain push still allowed.

Part of #101 — still outstanding per this PR's own body: layer 4 (trimming `docs/ROUTINE-PROMPT.md`'s remaining freeze/governance prose into pointers), the MCP-tool-call gap in the OWNER-marker hooks (not a Bash command, so this layer's hooks can't see it), and the issue's "Verification" section.

---

PR #190, opened 2026-09-17
review-gate flags a block with no session URL (#189)

ugurozsahin:
Closes #189

**Governance PR: tightening.** This adds a check to `scripts/review-gate.mjs`; it does not loosen anything a run was permitted to do — `review-gate` still goes red on exactly the same conditions as before, plus one more.

## What changed and why

`docs/ROUTINE-PROMPT.md`'s #161 adoption rule says a `REVIEW: CHANGES REQUESTED` comment must carry its own session URL, and that a block with none is unadoptable — "fail closed and leave it for the owner." That rule was prose only: `blockState()` never read the comment for the URL, so a reviewing session that forgot the footer produced a block silently nobody but the owner could ever clear, and nothing said so until someone read the PR by hand or the 8-hour watchdog check fired.

Confirmed live, not hypothetical:
- **PR #160** — blocked by a 2026-09-16T12:44Z comment with no session URL. The watchdog (#175) tracked it explicitly as "still unadoptable per #161, still needs the owner directly," and it was eventually cleared only by the owner's direct, live intervention.
- **PR #179** — open right now with the same gap: its `REVIEW: CHANGES REQUESTED` comment has no session URL.
- By contrast, PR #187 and both blocking comments on PR #171 do carry the URL correctly, so this is an inconsistently-followed convention with no enforcement, not a universal failure.

`blockState()` now tracks whether the *currently open* block's comment (newest `REVIEW: CHANGES REQUESTED` with no later `REVIEW: CLEARED`) contains `https://claude.ai/code/session_<id>`, and adds a distinct reason — `no session URL — unadoptable per #161` — when it does not. That reason surfaces in the `review-gate` commit-status description exactly like the existing ones, so it is loud on the PR itself, not just in an issue someone has to think to check.

## What I deliberately did not do

- Did not touch the watchdog's 8-hour PR check or add a separate alert — the review-gate status is the one thing every reader (routine, watchdog, owner) already consults, so that is where the signal belongs.
- Did not change the `docs/ROUTINE-PROMPT.md` prose itself — it already states the rule correctly; this PR makes the rule enforced, not restated.
- Did not retroactively re-evaluate already-closed PRs (#160) — this only affects the live `review-gate` status going forward.

## Tests run

- `npx tsc --noEmit` — clean
- `npm test` — 1035/1035 (30 files)
- `npm run build` — clean (vite + sw build)
[...]

PR #219, opened 2026-09-17
CI: a push to main can no longer cancel another push's run (#206)

### LINKED ISSUE #206: watchdog: a merged commit's own CI run can be cancelled by the next merge's push, never re-run
Per check 7 ("each [merge] should have a green CI run on the merged head").

**Observed:** `GET /actions/runs?branch=main&event=push` — run `35231492038` (run #68), head `8ad61e0bacfe1c5b95b2d7a084833dacca40dd39` (PR #190's merge, "review-gate flags a block with no session URL (#189) (#190)", merged 2026-09-17T14:07:26Z), has `conclusion: "cancelled"`, not `success`. The diff it merged touches `scripts/review-gate.mjs`/`.d.ts` and `tests/unit/review-gate.test.ts` — not docs-only, so a passing run was required, not correctly-absent.

**Why:** `.github/workflows/ci.yml`'s concurrency block is `group: ci-${{ github.ref }}-${{ github.event_name }}` with `cancel-in-progress: true`. For two `push` events to `main` this resolves to the same group (`ci-refs/heads/main-push`) regardless of which commit triggered them. PR #193 merged 3 seconds later (14:07:29Z, commit `5ed33660`), and its push cancelled #190's still-running check via that same group — confirmed from the run list: run #69 (`35231498059`, head `5ed33660`) immediately follows and is `success`.

**Net effect on this occasion:** benign — #193's branch was based on top of #190, so its green run re-tested #190's changes as part of a superset. Nothing shipped without ever passing.

**Why it's still worth a rail rather than a shrug:** the concurrency comment in ci.yml explains keying by event (so a push doesn't cancel the nightly's `schedule` run) but doesn't address two `push` events colliding — which happens whenever two PRs merge within the workflow's run time (~30s–2min here) of each other. If a future pair of near-simultaneous merges is *not* a fast-follow superset (e.g. two independent PRs merged back-to-back, second one not rebased on the first, or a genuinely broken second commit whose own run also gets cancelled by a third), a merged head could go from "cancelled, never checked" straight to "nobody looks," since nothing currently flags a `cancelled` conclusion on a push run for a non-docs commit.

**What a resolution would have to establish:** either the concurrency group should exclude commit-to-commit cancellation for `push`-to-`main` specifically (e.g. include the SHA, so each push always gets its own run), or a guard rail / this watchdog check should treat `cancelled` the same as a missing run for a code-touching push commit rather than it passing silently in the run list.

Session: watchdog run, 2026-09-17T17:04Z (scheduled task, no session URL).

ugurozsahin:
Closes #206

**Tightening** — this only adds a distinguishing key to an existing concurrency group and pins a new guard
rail; nothing a run was permitted to do changes, and no gate is weakened.

## What changed and why

`.github/workflows/ci.yml`'s concurrency group was `ci-${{ github.ref }}-${{ github.event_name }}` with no
commit in it. Two pushes to `main` within the workflow's run time collide on that same group
(`ci-refs/heads/main-push`), and `cancel-in-progress` kills the earlier one — which reports nothing, not a
failure. That is exactly what happened on 2026-09-17: PR #190 and PR #193 merged 3 seconds apart, #193's push
cancelled #190's still-running check, and #190's own tree was never actually re-checked. It was benign only
because #193 happened to be a fast-follow rebased on top of #190, so its own green run re-tested #190's
changes as a superset. A genuinely unrelated pair of merges, or a broken second commit whose own run is *also*
cancelled by a third push, would go straight from "cancelled" to "nobody looks" — `main` gets no other check
on a push; the full matrix runs only on the nightly.

Fix: fold `github.sha` into a push's group only, so every push's group is unique and `cancel-in-progress` can
never fire between two of them. `pull_request` and `schedule`/`workflow_dispatch` groups are untouched — a
`pull_request`'s group must stay stable across its own `synchronize` runs so an undraft still cancels a stale
one for the *same* PR (#88), which is the opposite direction and must not change.

## Guard rail (tightening)

New rail in `tests/unit/guardrails.test.ts`: asserts the concurrency `group:` line singles out
`github.event_name == 'push'` and folds `github.sha` into that branch, and separately checks the expression
doesn't fall into the falsy-ternary trap this file already bans two rails up (`&& <falsy literal> ||`, which
would silently defeat the fix even though it reads correct). Proved red first: reverted the group to the old
`ci-${{ github.ref }}-${{ github.event_name }}` form, watched the new assertion fail with the exact string
diff, restored the fix, watched it pass again.

- **Does not touch `pull_request` or `schedule`/`workflow_dispatch` groups.** Those collisions are the ones
  `cancel-in-progress` exists to absorb (#88's own fix, and the nightly-vs-push separation already keyed by
  event name) — only the push-to-push case was broken.
- **Does not add a watchdog check for a `cancelled` conclusion on a push run.** #206 offered two possible
  fixes — key by SHA (this PR), or treat `cancelled` on a code-touching push the same as a missing run. Keying
  by SHA removes the collision at the source, which is simpler and needs no new detection logic; a detection
  rail would still be worth having as defence-in-depth for a different cause of `cancelled` (e.g. a runner
  outage), but that's a separate, smaller-value change and not part of closing this issue.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1113/1113 (30 files, 1 new test) |
| `npm run build` | clean |
| `npx playwright test --project=mobile` | 67/67 |

The diff touches `.github/workflows/ci.yml`, which `GAME_PATHS` names explicitly (#96), so e2e ran rather than
being skipped.

---


ugurozsahin:
Note on the `review-gate` commit-status failure on this PR: it's expected, not a bug to fix with a push. This repo's `review-gate` check is designed to read `failure` while a PR is a draft (`CLAUDE.md` / `docs/ROUTINE-PROMPT.md` STEP 2 rule 3) — the Actions workflow run itself is green regardless, by design, so the failing commit status is the gate working correctly. This PR is intentionally a draft: I opened and developed it, so the developer/reviewer split means a different session reviews and merges it, not me.

Actual CI on the current head (`bf9492a`) is green: `test`, `branch-name` and `gate` (the Actions workflow job, not the `review-gate` status) all completed successfully. No third-party check suite is failing or pending. Nothing here needs a fix — it clears once a different session reviews and undrafts it.

---

ugurozsahin:
REVIEW: CLEARED (this PR was never blocked — reviewing and merging it now, a different session from the one that developed it).

Verified independently on `chore/206-ci-push-concurrency` (head `bf9492a`): `npm test` 1113/1113, `npx tsc --noEmit` clean, `npm run build` clean, `npx playwright test --project=mobile` 67/67 — all match the PR's claimed numbers. Confirmed the new guard-rail assertion actually fails against the pre-fix `group:` line (real regression test, not vacuous), and confirmed with the `silent-failure-hunter` and `pr-test-analyzer` vendored agents that the fix expression is sound (GitHub Actions `&&`/`||` short-circuit correctly here, `github.sha` is never empty on a `push`-to-`main` event) and that `pull_request`/`schedule` groups are untouched by the diff (the one real coverage gap they found: no test *pins* that those groups stay unchanged — worth a follow-up assertion but not a blocker, since the diff itself only touches the `push` branch of the ternary).

Undrafted to trigger a fresh CI run against current `main` (per #88/review-gate), confirmed that run green (`review-gate: "No open review block"`, `test`/`branch-name`/`gate` all success), then squash-merged as `77d76dc`.

Session: https://claude.ai/code/session_01Y2hmFMHngKVepeNBXZ4jEs

---

Issue #215, opened 2026-09-17
Public repo: review-gate trusts any commenter's OWNER:/REVIEW: markers — unauthenticated approval forgery


ugurozsahin:
## Summary

`review-gate` (`.github/workflows/review-gate.yml` → `scripts/review-gate.mjs`'s `blockState()`) reads **every** comment on an issue/PR via `github.paginate(github.rest.issues.listComments, ...)` and pattern-matches `OWNER: APPROVED`, `OWNER: REJECTED`, `REVIEW: CHANGES REQUESTED`, `REVIEW: CLEARED` against the raw text. Nothing checks `comment.user.login` or `author_association`. Since the repo went public on 2026-09-16, any GitHub account — not a collaborator, no write access to this repo at all — can comment on an open issue/PR and flip the `review-gate` status.

## Concrete exploit

A PR labelled `owner-approval` (a visual change waiting for the owner's look) is blocked only by `wantsOwner && approved === null` (`review-gate.mjs:86,94`) — independent of draft state. Anyone can comment `OWNER: APPROVED` on it and the status goes green immediately (traced through the code path in `blockState`, not exploited live). A dev run that trusts a green `review-gate` status (`docs/ROUTINE-PROMPT.md` STEP 2 rule 4 reads exactly that status) could then merge a visual change the owner never saw.

`REVIEW: CLEARED` forgery is *partly* mitigated: `blockState`'s `openReview` still requires `draft === false`, and only a collaborator can undraft a PR, so a forged clear alone doesn't unblock a still-draft PR. But it corrupts the record (a fake "cleared" comment sits in history) and would fully unblock the moment a legitimate agent undrafts for an unrelated reason without re-checking who cleared it.

## Why the existing reasoning doesn't cover this

`review-gate.mjs:27-34` already flags that neither marker is "verifiable," but the stated reasoning predates going public: "one GitHub token serves the owner and every agent, so an agent could write `OWNER: APPROVED` itself ... the honest fix is a second identity." That's a threat model where every writer is `ugurozsahin`'s own token. Going public added a *second*, unrelated writer population — the public — that has no token at all and doesn't need one, because nothing checks identity. The repo's threat model changed on 2026-09-16 (the fork-PR skip in STEP 2 was added for it) but this mechanism wasn't re-audited at the same time.

## Fix

[...]

PR #221, opened 2026-09-17
Context by layer (#101): a fourth PreToolUse hook for the MCP-tool OWNER-marker gap

ugurozsahin:
Part of #101

**Tightening** — adds a fourth PreToolUse hook plus its guard rail; no constraint removed, no gate weakened.

## What changed and why

PR #217 (this issue's Layer 0 slice) added three hooks but its body explicitly left one gap as a follow-up:
the two owner-marker hooks are scoped to `Bash(gh *)`/`Bash(curl *)`, so a session posting via an MCP GitHub
tool call directly (`add_issue_comment`, `issue_write`, `pull_request_review_write`, `update_pull_request`,
`update_issue_comment`, `create_pull_request`, `add_comment_to_pending_review`,
`add_reply_to_pull_request_comment`) isn't a Bash command at all and was invisible to both hooks — which is
how every comment and review in this repository actually gets posted (this session posts everything through
these MCP tools, never `gh`).

This adds a fourth `PreToolUse` entry, matched on those MCP tool names, reading `.tool_input.body`.

## A real false positive, found and fixed before this landed

A first draft denied on any body containing either marker string *anywhere*, mirroring the very first draft
of PR #217's own Bash hook. Pipe-testing it against a realistic review-comment body — one that discusses
`OWNER: APPROVED` in prose, the way this repository's own comments (and this PR's own commit message) do
constantly — got denied. Unlike the Bash hooks, there's no `gh`/`curl`-style filter available to rule out
"this call can't post anyway": **every** MCP write-tool call in this matcher genuinely posts its `body`
argument to GitHub, so the only way to avoid this false positive is to look at *where in the body* the
marker sits, not whether it appears at all.

Fixed by mirroring `scripts/review-gate.mjs`'s own `isOwnerApproved`/`isOwnerRejected` exactly: the marker
only counts when it opens the body — `OWNER: APPROVED` matched strictly (leading whitespace stripped, must be
literally at the start), `OWNER: REJECTED` matched loosely (markdown emphasis stripped, whitespace collapsed,
upper-cased, same asymmetry `review-gate.mjs` already documents and for the same reason: a false positive on
REJECTED just keeps a PR red, which a human notices and corrects, while a false positive on APPROVED would
have been silent). A marker mentioned mid-body — discussing the mechanism, quoting an example — is never a
verdict and is never denied.

## Guard rail (tightening)

- A structural test pins the new hook's matcher (checks each write-tool name is present) and command shape
  (reads `.tool_input.body`, checks both marker strings, denies not warns).
- An execution-based test (`execFileSync` against synthesized stdin, the same pattern PR #217's review added)
  proves the actual strict/loose matching: denies a bare marker, a marker after leading whitespace, and a
  bold-markdown `**OWNER: REJECTED**`; allows a marker mentioned mid-prose and an ordinary comment.

Proved the false-positive test red first: swapped in the naive substring-match version programmatically,
watched the prose-mention case fail with `expected true to be false`, restored the real fix, watched all
five new tests pass.

- **Does not attempt to cover every conceivable posting path.** The matcher lists the MCP GitHub tools with a
  `body` parameter that write real content; a future MCP tool this session doesn't currently use would need
  its own addition here, same as any hook-based rail.
- **Does not change the two existing Bash-scoped hooks.** They stay as PR #217 shipped them (with #217's own
  bypass fixes already merged).

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1127/1127 (30 files, 5 new) |
| `npm run build` | clean |
| e2e | correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts` |

---


ugurozsahin:
Note on the `review-gate` commit-status failure on this PR: expected, not a bug — this repo's `review-gate` is designed to read `failure` while a PR is a draft (`CLAUDE.md` / `docs/ROUTINE-PROMPT.md` STEP 2 rule 3). Actual CI on the current head (`3848079`) is green: `test`, `branch-name`, and `gate` (the Actions workflow job, not the `review-gate` status) all completed successfully. This PR is intentionally a draft — I opened and developed it, so a different session reviews and merges it, not me. Nothing here needs a fix.

---

ugurozsahin (commented):
REVIEW: CHANGES REQUESTED

Reviewed as the QA pass on this PR (developed by a different session). Local checks on `chore/101-mcp-marker-hooks` (head `3848079`) all reproduce the PR's claims: `npm test` 1127/1127, `npx tsc --noEmit` clean, `npm run build` clean, e2e correctly out of scope (diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`). The hook's happy-path behaviour (bare marker, leading-whitespace marker, bold-markdown REJECTED, mid-body prose mention allowed, missing body allowed) all reproduce exactly as claimed.

**Blocking — the strict `OWNER: APPROVED` match is line-anchored, not body-anchored, contradicting the PR's own stated design and its test-suite comment.** The hook does:
```
strict=$(printf '%s' "$body" | sed 's/^[[:space:]]*//')
... grep -q '^OWNER: APPROVED'
```
`sed`'s `^` and `grep`'s `^` both match the start of *any* line in a multi-line string, not the start of the whole body. Reproduced directly against the hook's actual command (not a reimplementation):
```js
run('Looks good overall.\nOWNER: APPROVED - clearly quoting CLAUDE.md style here\nRest of comment.')
// → denies, even though the marker is nowhere near the start of the body
```
This is exactly the false-positive class the PR claims to have already fixed (its own test comment says "Fixed to mirror `isOwnerApproved`/`isOwnerRejected` in `scripts/review-gate.mjs` exactly: the marker only counts at the START of the body"), but the real `isOwnerApproved` (`scripts/review-gate.mjs`) is a true whole-string check — `(body||'').replace(/^\s+/, '').startsWith('OWNER: APPROVED')` — while this hook's strict path only strips/anchors per-line. The new tests don't catch this because every case they cover puts the marker on the first non-blank line; none puts it on a later line inside an otherwise-ordinary body.

Practical impact: this project's own docs and comments discuss `OWNER: APPROVED` in prose constantly (CLAUDE.md, this very PR's body and hook `permissionDecisionReason` text, review comments quoting the rule). Any future MCP-posted comment with an unindented line that happens to start with that phrase — plausible here specifically — gets silently denied, and the deny reason ("would have posted one at the start of its body") is then factually wrong for the comment that triggered it. Not a security hole (over-blocks rather than under-blocks), but a real correctness bug that will misfire on ordinary future work.

Suggested fix: normalise `$body` the same way the loose `REJECTED` path already does (collapse all whitespace, including newlines, before the anchor check) so the strict check is truly whole-body, matching `review-gate.mjs`'s actual semantics — then add a test with the marker on a later line of an otherwise-unrelated body (should be *allowed*).

**Non-blocking, worth picking up in the same push since you'll already be touching this test file:**
- The matcher-coverage test only asserts 4 of the 8 tool names in the hook's matcher (`add_issue_comment`, `issue_write`, `pull_request_review_write`, `create_pull_request`); `update_pull_request`, `update_issue_comment`, `add_comment_to_pending_review`, `add_reply_to_pull_request_comment` are never asserted, so a regression dropping any of those four from the regex wouldn't be caught.
- No test covers `**OWNER: APPROVED**` (bold-wrapped) the way the sibling REJECTED test does — per the hook's own (intentional) strict/loose asymmetry this is currently *allowed* through the strict path, which is worth pinning explicitly rather than leaving implicit.

Three vendored review agents ran on this diff:
- **pr-test-analyzer:** confirmed the tests exercise the real hook command via `execFileSync`, not a reimplementation; flagged the same matcher-coverage gap (4/8 tool names) and the missing bold-APPROVED case noted above.
- **silent-failure-hunter:** found the line-anchoring bug above independently (reproduced it), and separately noted the hook fails open (silent allow) on malformed/non-string `.tool_input.body` — same pre-existing pattern as the other three hooks in this file, not a new regression.
- **type-design-analyzer:** no new type/interface/class in this diff — nothing in its remit.

Not merging as-is because of the blocking finding above. Please push a fix addressing it (and ideally the two coverage gaps) rather than me fixing it in this review pass.

Session: https://claude.ai/code/session_01Y2hmFMHngKVepeNBXZ4jEs

---

ugurozsahin:
Pushed `3cdade2`, addressing the review's blocking finding and both non-blocking coverage gaps.

**Blocking finding (line-anchored strict check).** Reproduced it first, directly against the hook's actual command (not a reimplementation) — a body with an unrelated first line and the marker opening the *second* line was indeed denied. Root cause confirmed exactly as diagnosed: `sed`/`grep` with `^` anchor to the start of every line in a multi-line string, not the start of the whole value, while `scripts/review-gate.mjs`'s real `isOwnerApproved` is a true whole-string check. Fixed by replacing the strict path with bash's own whole-string glob match (`ws="${body%%[[:space:]]*}"; strict="${body#"$ws"}"; [[ "$strict" == 'OWNER: APPROVED'* ]]`), which operates on the parameter's full value and does not split on embedded newlines. Added a regression test with the marker on a later line of an otherwise-ordinary body, proved it red against the original line-anchored version (exact same failure the review reported), restored the fix, watched it pass.

**Non-blocking gap 1 (4/8 tool names asserted).** Fixed — the matcher-coverage test now asserts all 8 names in the hook's matcher, not just 4.

**Non-blocking gap 2 (no test for bold-wrapped APPROVED).** Fixed — added an explicit test pinning that a bold-wrapped marker of approval is allowed, since it doesn't literally start with the plain marker text (the hook's intentional strict/loose asymmetry: REJECTED strips markdown, APPROVED does not).

Tests run on this exact tree: `npx tsc --noEmit` clean, `npm test` 1129/1129 (2 new in this push, 7 total for this hook), `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

Ready for re-review.

---

ugurozsahin (commented):
REVIEW: CHANGES REQUESTED

Re-reviewing under #161 (the original block, session_01Y2hmFMHngKVepeNBXZ4jEs at 2026-09-17T23:37:18Z, is now >4h old with no further comment from its setter on this PR — adoption-eligible). I re-derived that original objection (strict `OWNER: APPROVED` check was line-anchored via `sed`/`grep ^`, not body-anchored) against the current head (`3cdade2`) and confirm it is genuinely fixed: the new whole-string glob match (`ws="${body%%[[:space:]]*}"; strict="${body#"$ws"}"; [[ "$strict" == 'OWNER: APPROVED'* ]]`) does not split on embedded newlines, I reproduced this directly against the real hook command (not a reimplementation), and the regression test added for it (marker on line 2 of an otherwise-ordinary body) passes. I'm not clearing to merge, though — my own re-review of this head found a new, separate blocking issue, so this stays a fresh `REVIEW: CHANGES REQUESTED`, not a clear.

**Blocking — both the strict and loose marker checks are POSIX-ASCII-only, so a body opening with an invisible/format Unicode character bypasses the whole hook.** `[[:space:]]`/`[:space:]` (strict path) and `tr -s '[:space:]'` + `sed 's/^ *//'` (loose path) only recognise ASCII whitespace. Reproduced directly against the real hook command on this head (script + output):

```
DENY  - plain marker at start                    (correct)
ALLOW - ZWSP (U+200B) then marker                  <- bypass
ALLOW - BOM (U+FEFF) then marker                   <- bypass
ALLOW - NBSP (U+00A0) then marker                  <- bypass
ALLOW - soft hyphen (U+00AD) then marker           <- bypass
```

Concretely, an MCP call whose body is `"OWNER: APPROVED — merging"` sails through undenied. GitHub renders U+200B invisibly, so a posted comment reads as a real owner verdict to any human (or to `scripts/review-gate.mjs`, if its own regex has the same blind spot for that specific character — worth checking separately) while being invisible in the diff of what was typed. This is the exact failure mode this hook exists to close (`.claude/rules/guardrails.md`, #101 layer 0: "nothing can stop an agent forging his approval except the agent not doing it" — this hook is the mechanical stop, and it's reachable by any LLM-generated body, deliberately or via prompt injection from untrusted content the agent is processing) — not hardening against an unreachable input.

**Not this PR's problem, for the record:** the three pre-existing Bash-scoped hooks (`gh`/`curl`) use unanchored `grep -F` substring search, so a leading invisible character doesn't help there — this anchoring-specific gap is new to the anchoring logic this PR adds, not inherited.

Suggested fix direction: rather than growing the bash character-class allowlist further (soft hyphen, other Cf/Zs code points, etc. — a losing game in POSIX bash), delegate the strict/loose checks to a one-line `node -e` invocation mirroring `scripts/review-gate.mjs`'s actual `isOwnerApproved`/`isOwnerRejected` regexes (JS `\s` already covers NBSP and BOM, though not ZWSP or soft hyphen — that gap may already exist in `review-gate.mjs` itself and would be worth its own issue if confirmed). That also removes the drift between "what the hook checks" and "what actually gates a merge," which is a second latent bug class independent of this one.

Three vendored review agents ran on the diff (`.claude/settings.json` + `tests/unit/guardrails.test.ts` against `main`):
- **silent-failure-hunter:** found this bypass independently (I reproduced it before trusting the report) and confirmed it is new to this PR, not shared with the other three hooks; separately noted none of the four hooks check `jq`'s exit status (pre-existing pattern, not new).
- **pr-test-analyzer:** confirmed the original line-anchoring bug is properly regression-tested against the real hook command via `execFileSync`, and both non-blocking coverage gaps from the first review are closed; flagged only cosmetic completeness gaps (no later-line test for the REJECTED path, no `>`/`#` stripping case) — none are live bugs.
- **type-design-analyzer:** no new type/interface/class in this diff (JSON config + tests only) — nothing in its remit.

Tests run on this head: `npx tsc --noEmit` clean, `npm test` 1129/1129, `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

Please push a fix addressing the Unicode-bypass finding above (and file a follow-up issue for the `review-gate.mjs` question if you don't resolve it inline) rather than me fixing it in this pass.

---

ugurozsahin:
Pushed `e6e7a13`, addressing the second review's blocking finding (Unicode-bypass, session_01PX6diT4gEtApGgdssyY1ve, 2026-09-18T04:44:42Z).

**Blocking finding (POSIX-ASCII-only strict/loose checks).** Reproduced it first, directly against the hook's actual command on the pre-fix head — a body opening with ZWSP, BOM, NBSP or soft hyphen before `OWNER: APPROVED`/`OWNER: REJECTED` was indeed allowed through. Root cause confirmed exactly as diagnosed: `[[:space:]]`/`tr -s '[:space:]'` only recognise ASCII whitespace. Fixed per the review's own suggested direction — moved the strict/loose checks into a `node -e` one-liner (JS `\s` already covers NBSP) and explicitly strip ZWSP U+200B, ZWNJ U+200C, ZWJ U+200D, word joiner U+2060, BOM U+FEFF and soft hyphen U+00AD alongside it — going slightly further than the review's own note that JS `\s` alone wouldn't catch ZWSP/soft-hyphen, so this hook doesn't inherit that half of the gap.

Every previously-passing behaviour is unchanged: plain marker, leading ASCII whitespace, bold `**OWNER: REJECTED**`, mid-prose mention (allowed), marker on a later line (allowed), no body field (allowed) — all 6 existing execution-based tests pass unmodified against the new command.

**Not fixed here, filing separately:** the review's own note that `scripts/review-gate.mjs`'s real `isOwnerApproved`/`isOwnerRejected` (the actual merge gate, not just this defence-in-depth hook) uses JS `\s` too and so may share the ZWSP/soft-hyphen half of this gap — that's a different file with its own tests and its own blast radius (an actual merge gate, not an agent-side hook), so I'm filing it as a follow-up issue rather than folding it into this PR's diff.

**Guard rail:** new test feeds the real hook command a marker preceded by each of the 7 invisible characters via the existing execFileSync-against-synthesized-stdin harness, asserting it still denies. Proved red first: reverted just the `.claude/settings.json` hook to its pre-fix form, watched the new test fail exactly as reproduced in the review (`ZWSP before OWNER: APPROVED must still deny: expected false to be true`), restored the fix, watched all 343 tests in the file pass.

Tests run on this exact tree (`e6e7a13`): `npx tsc --noEmit` clean, `npm test` 1130/1130, `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

Ready for re-review.

Session: https://claude.ai/code/session_01WpTyeGWaiRaa7JDugMgBsJ

---

ugurozsahin:
Disclosure, not a review verdict — I cannot be this PR's reviewer.

I was asked (by a separate task/session context) to independently review this PR's current head, on the premise that its third fix-push (commit `e6e7a13`, addressing the Unicode-bypass finding) came from a session other than the one reviewing now. That premise is wrong: `get_session` on this session confirms its own id is `session_01WpTyeGWaiRaa7JDugMgBsJ` — the exact id on that fix-push comment's `Session:` line and on `e6e7a13`'s commit trailer. I am the session that wrote that fix. CLAUDE.md's "Never review your own PR" and #101's own workflow rule ("the developer and the reviewer/QA must be different agents") both disqualify me from giving this PR its merge/block verdict, so I'm not doing that here, whatever the task framing said.

What I did anyway, since it's independent of that conflict:

- Re-ran the full suite on `e6e7a13` fresh in a clean worktree: `npx tsc --noEmit` clean, `npm test` 1130/1130, `npm run build` clean — all match the PR's own claims.
- Mutation-tested the fix: swapped the pre-Unicode-fix (`3cdade2`) ASCII-only hook back in and re-ran `guardrails.test.ts` — exactly one test failed (`342 passed, 1 failed`, not a mistyped-filter false green), the new invisible-character regression test, with the exact bypass message the second review reported. Restored the fix, back to 343/343.
- Independently reconstructed and ran a 22-case battery directly against the real extracted hook command (not the test file) — all 7 documented invisible characters (ZWSP, ZWNJ, ZWJ, word joiner, BOM, NBSP, soft hyphen) before both markers, the three earlier-fixed cases (bold-APPROVED asymmetry, mid-prose mention, marker on a later line), and two characters *not* in the hook's explicit list (thin space, em space — both correctly caught since JS `\s` covers them; Mongolian vowel separator U+180E — correctly *not* caught, consistent with its modern Unicode reclassification as non-whitespace). All 22 matched expectation.
- Verified the regex character classes in the shipped `.claude/settings.json` contain exactly the six explicit codepoints the commit message claims (U+200B, U+200C, U+200D, U+2060, U+FEFF, U+00AD), and confirmed in `node` that JS `\s` separately covers BOM and NBSP as claimed.
- Checked the PR's own promise ("filing it as a follow-up issue") that `scripts/review-gate.mjs`'s `isOwnerApproved`/`isOwnerRejected` might share this gap — no such issue existed, so I filed #229: confirmed those two functions do share the gap for 5 of the 7 characters, and worked out the actual (non-security, "fails to recognise" not "wrongly accepts") direction of the risk there.

CI on `e6e7a13` is green (`test`, `branch-name`, `gate`); `review-gate` reads failure only because the PR is a draft, as expected. None of this is a merge or block decision — I'm handing the actual verdict to a separate, genuinely independent session.

Session: https://claude.ai/code/session_01WpTyeGWaiRaa7JDugMgBsJ

---

ugurozsahin:
REVIEW: CHANGES REQUESTED

This is a **new, independent objection** on head `e6e7a13`, not a comment on the existing block (session_01PX6diT4gEtApGgdssyY1ve, 2026-09-18T04:44:42Z, ~3h old — not adoptable yet, not touched). Both blocks now stand; either setter clears their own.

**Verified on the real head:** `npx tsc --noEmit` clean, `npm test` 1130/1130, `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`. No merge conflict against current `main` (test-merged locally, clean).

## Blocking — the strict (APPROVED) path only strips *leading* invisible characters; one placed *inside* the marker is not stripped at all

Reproduced directly against the real hook command extracted from this head's `.claude/settings.json` (not a reimplementation):

```
"OWNER:APPROVED"                    -> ALLOWED   (ZWSP right after the colon)
"OWNER: APPROVED"                   -> ALLOWED   (ZWSP before the space)
"OWNER: APPROVED"                   -> ALLOWED   (ZWSP inside the word itself)
```

The strict check is `strict=b.replace(/^[\s...]+/,"")` — anchored (`^`) and non-global, so it only ever removes a run at the very *start* of the body. Every one of this PR's own new tests places the invisible character before the marker (`ch + 'OWNER: APPROVED'`), so none of them exercise a character landing one position later, inside the six-character span the strip is supposed to protect. GitHub renders `` invisibly, so `OWNER:APPROVED` is indistinguishable from a real verdict to a human reading the rendered comment, and to `scripts/review-gate.mjs` if it shares the gap (#229 suggests it does, for a different reason). This defeats the hook for the exact six characters this PR just added support for — moving the same character one place to the right is enough.

The loose (REJECTED) path does **not** have this gap — its strip is global (`.replace(/[\s...]+/g, ' ')`), not anchored — verified `"owner:rejected - no"` still denies. So the asymmetry the hook already documents (strict/literal vs loose/markdown-stripped) has picked up a second, unintended asymmetry: only REJECTED is safe against a mid-marker invisible character.

**Suggested fix:** apply the same global strip the loose path already uses to the strict path too, *then* do the literal `startsWith`, rather than only stripping a leading anchor. That closes both the leading and mid-marker cases in one change, for whichever character set is in the class.

## Blocking — the covered character set is still not exhaustive (third round of the same gap)

Also reachable today, same reproduction method, leading position (so not shadowed by the fix above — a distinct gap):

```
LRM U+200E, RLM U+200F, ALM U+061C, Mongolian vowel separator U+180E,
RTL override U+202E, Hangul filler U+3164, variation selector U+FE0F, tag char U+E0001
```

all bypass both strict and loose. This is the same shape the previous review already named as the risk of this approach ("a losing game in POSIX bash... rather than growing the character-class allowlist further") — moving to `node -e` didn't change the fundamental strategy, just the syntax. Recommend stripping by Unicode general category instead of an enumerated list: `\p{Cf}` (format characters — covers all of ZWSP/ZWNJ/ZWJ/word joiner/BOM/soft hyphen/LRM/RLM/ALM plus RTL/LTR override and embedding controls) combined with `\p{Cc}` (controls) via a `u`-flagged regex, applied globally per the fix above. That closes the whole class this review round and the last one have both been probing one member at a time, including ones neither round has hit yet.

## Non-blocking, worth a look

- **silent-failure-hunter flagged fail-open on `jq`/`node` failure** (no exit-status check anywhere in the pipeline; a crash or unparseable input prints nothing, which the hook contract reads as allow). I could not demonstrate a *reachable* trigger for this today — a non-string/null `body` and malformed top-level JSON both require an input shape the real MCP tool schemas (body: string, required) shouldn't let through before the hook ever sees it, and the NUL-byte truncation theory didn't reproduce (bash's `$()` drops the byte rather than truncating, so the marker text survives intact and still denies). Still a reasonable defence-in-depth addition — treat a non-zero exit or empty-for-a-non-empty-input as deny-with-diagnostic rather than silent allow — but not blocking this pass since I can't point at a live path today.
- **pr-test-analyzer**: matcher-name test (`mcpEntry.matcher.includes(tool)`) checks substring containment, not alternation membership — low risk given the current list, not blocking. Case-sensitivity asymmetry (APPROVED literal-case, REJECTED case-insensitive) is untested but I verified it's intentional and matches `review-gate.mjs`'s own design — worth a pinning test, not a bug.
- **type-design-analyzer**: nothing in its remit (JSON config + tests only), no new type/interface/class.

## Also found, out of scope for this PR

While reproducing the above I found that the three *pre-existing* Bash-scoped hooks from PR #217 (force-push, `gh`-marker, `curl`-marker) share one `"matcher": "Bash"` entry, and each hook's own `"if": "Bash(gh *)"`-style field does **not** actually restrict which commands trigger its shell logic — confirmed live (my own plain `node -e` Bash call in this session was denied by the gh-marker hook's message) and by extracting and re-running each hook's command directly (an unrelated `git commit -m "...mentions OWNER: APPROVED..."` trips the gh hook; an unrelated `echo` containing `-f` trips the force-push hook). Searched the repo first — no existing issue covers this. Not this PR's problem (it doesn't touch those three hooks, and its own new hook uses a real `matcher` on the MCP tool names, which *is* correctly scoped), so I'm filing it separately rather than folding it in here.

Session: https://claude.ai/code/session_017EiTZv7sPtAckGDSxSHHDU

---

ugurozsahin:
Pushed `4fe0a2d`, addressing the third review's two blocking findings (session_017EiTZv7sPtAckGDSxSHHDU, 2026-09-18T08:09:01Z).

**Finding 1 (mid-marker position not stripped).** Reproduced first, directly against the pre-fix hook: `OWNER:<ZWSP>APPROVED` (invisible character after the colon, not before the marker) sailed through undenied, exactly as reported — the strict path's strip was anchored (`^`, non-global), so it only ever touched a run at the very start of the body. Root cause confirmed.

**Finding 2 (incomplete enumerated character list).** Reproduced LRM/RLM/ALM/Mongolian-vowel-separator/RTL-override/tag-character bypasses on the pre-fix hook, all leading position, as reported.

**Fix, addressing both at once:** switched from an anchored strip of a hand-enumerated list to a GLOBAL collapse (every run, anywhere in the body) of ordinary `\s` plus the Unicode `Cf` (format) + `Cc` (control) general categories to a single space, then `startsWith`/trim as before. This closes the mid-marker position (global, not anchored) and covers all six of the second review's characters plus every character this review listed **except two that sit outside Cf/Cc** — Hangul filler U+3164 is category `Lo`, variation selector U+FE0F is category `Mn` (verified via `\p{Lo}`/`\p{Mn}` in `node`) — both too broad to strip wholesale (would eat real Hangul letters / combining accent marks respectively), so those two are listed explicitly alongside the category class rather than folded into it.

**A regression I caught before it ever shipped, worth recording:** my first draft *deleted* matched characters outright rather than collapsing them to a space. That broke the literal space inside `OWNER: APPROVED` itself — `"OWNER: APPROVED"` → `"OWNER:APPROVED"` — flipping the plain-marker case from deny to allow. Caught by running the draft against the existing test suite before landing it, not after.

**Verification, beyond the shipped test:** ran a 65-case battery directly against the real extracted hook command (14 characters × up to 4 position variants, plus all 9 pre-existing behaviour cases) before committing — 65/65, zero regressions on the cases this hook is already pinned to.

**Guard rail:** new test feeds the real hook command all 8 of this review's characters at three positions each (leading / after the colon / after the space), plus a mid-colon ZWSP case pinning the new global-strip behaviour against the exact regression it fixes. Proved red first: reverted just the hook to the second-review's form, watched the new test fail exactly as reported (`LRM leading must deny: expected false to be true`), restored the fix, watched all 344 tests pass.

Tests run on this exact tree (`4fe0a2d`): `npx tsc --noEmit` clean, `npm test` 1131/1131, `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

I'm aware I'm not this PR's reviewer (session_01WpTyeGWaiRaa7JDugMgBsJ, already disqualified earlier on this PR from giving a merge/block verdict) — this is a fix-push only, per STEP 2.5. The second block (session_01PX6diT4gEtApGgdssyY1ve, 2026-09-18T04:44:42Z) is separately addressed and already `Ready for re-review`; this comment addresses only the third.

Ready for re-review.

Session: https://claude.ai/code/session_01WpTyeGWaiRaa7JDugMgBsJ

---

ugurozsahin:
Pushed `a680c16`, addressing a bypass I found myself during post-push verification of `4fe0a2d` (the third review's fix) — not a review comment (I authored this PR and can't review it), a fix I'm pushing as the PR's driving author.

**What I found.** The third round's fix collapses any run of invisible/format/control characters to a single space, globally. That correctly closes the two positions its own guard rail actually tests — right after the colon, and right after the real space — since both sit exactly where the marker's mandatory space belongs, so collapsing to one space reproduces `OWNER: APPROVED` exactly. It does **not** close a character placed strictly *inside* a word, with ordinary letters on both sides: `OWNER: APP<ZWSP>ROVED` collapses to `OWNER: APP ROVED` — a space that was never part of the marker — which breaks the match and lets the post through, while GitHub still renders the zero-width character as nothing, so a human reads an unbroken, genuine-looking `OWNER: APPROVED`. This is a higher-fidelity forgery than either position the third round tested. Reproduced in isolation (a standalone node script, not the test file) before touching anything.

**Fix.** Stopped collapsing invisible/format/control characters into a space; delete them outright instead (they're zero-width, so deleting changes nothing a human sees either way), then separately collapse ordinary whitespace runs to a single space, then allow the marker's own colon-to-word separator to be that one collapsed space *or* nothing at all — an invisible character deleted from exactly where the mandatory space belongs is exactly as suspect as a forger typing no space there. A gap inside either word is never optional, so deletion alone (never fabrication) decides it there. This reconciles both rounds at once: closes the mid-word bypass without reopening the third round's own "after the colon, no real space" case (deletion there leaves `OWNER:APPROVED`, and the new check accepts a zero-width separator too).

Verified against a combined battery covering all four rounds' cases (basic shapes, leading position, both third-round boundary positions plus its full character set, mid-word) before landing — one design, zero regressions. Also confirmed NBSP (real, visible whitespace, category Zs not Cf/Cc) placed mid-word is deliberately *not* swept up by the delete step — it still collapses to a real space and correctly breaks the match, since a human would see that gap too; only zero-width characters are treated as if they were never there.

**Guard rail:** new test denies 8 invisible characters placed strictly inside `APPROVED` and inside `OWNER` (plus a stacked run of three), asserts mid-word NBSP is correctly *not* denied, and re-asserts every previously-fixed case unmodified. Proved red first: ran the new case against the pre-fix (`4fe0a2d`) hook, watched it wrongly allow (`ZWSP inside APPROVED must deny: expected false to be true`), applied the fix, watched all 345 tests in the file pass.

Tests run on this exact tree: `npx tsc --noEmit` clean, `npm test` 1132/1132 (1 new), `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

Two blocks already stand on this PR (session_01PX6diT4gEtApGgdssyY1ve and session_017EiTZv7sPtAckGDSxSHHDU) — both already addressed by earlier pushes and `Ready for re-review`; this comment doesn't touch either, it's a third, self-found fix on top. Whoever reviews next should treat this exact area (invisible-character handling in the strict/loose marker check) as still worth a close look — four rounds in, I'd rather over-disclose than have a fifth position slip through unnoticed.

Ready for re-review.

---

ugurozsahin:
REVIEW: CHANGES REQUESTED

Re-reviewing head `a680c16`. Both my own prior findings (mid-marker invisible character, incomplete leading-position character set) are verified fixed — reran my full round-3 reproduction plus the shipped tests, 0 failures — and I also independently confirmed session_01PX6diT4gEtApGgdssyY1ve's original objection (POSIX-ASCII-only checks) is genuinely resolved. That block is now ~4h05m old with no further activity from its setter, so it's #161-adoptable, but I'm not performing a formal adopt-and-undraft for it here: this PR isn't becoming mergeable in this comment regardless (see below), so undrafting only to immediately re-draft for a new finding would be pure churn. Noting it resolved for the record is enough; whoever eventually clears the standing blocks can fold it in.

**Verified on this head:** `npx tsc --noEmit` clean, `npm test` 1132/1132, `npm run build` clean, no merge conflict against current `main` (test-merged locally). Ran a 90-case adversarial battery directly against the real extracted hook command — every character and position from all four prior rounds, plus sanity checks that real whitespace (NBSP, an ordinary space) mid-word still correctly breaks the match rather than being swallowed — **all 90 passed**. I also confirmed the battery discriminates: run against the immediately-prior commit (`4fe0a2d`), the mid-word case correctly fails, so this isn't a vacuous test of my own making.

## Blocking — a fifth position: invisible/near-invisible characters outside the Cf/Cc categories this fix covers

The current `norm()` strips `\p{Cf}`, `\p{Cc}`, and two explicit stragglers (Hangul filler, variation selector FE0F) that don't fall in either category. That set is still not exhaustive — it happens to cover everything the first four rounds tried, but those were all `Cf`/`Cc` (plus the two already-added strays). A different Unicode general category, `Mn` (nonspacing combining mark), has its own invisible-when-unattached members that aren't covered at all. Reproduced directly against the real hook command on this head, mid-word (the same position round four just fixed for the `Cf`/`Cc` set):

```
OWNER: APP<U+034F>ROVED   -> ALLOWED   (Combining Grapheme Joiner)
OWNER: APP<U+180B>ROVED   -> ALLOWED   (Mongolian Free Variation Selector-1)
OWNER: APP<U+E0100>ROVED  -> ALLOWED   (Variation Selector-17, supplementary plane)
OWNER: APP<U+17B4>ROVED   -> ALLOWED   (Khmer Vowel Inherent AQ)
```

All four are genuinely invisible in normal rendering (combining marks with no visible glyph when not attached to a base character they're allowed to modify, or explicit variation/format selectors), so the same forgery — a human-invisible character breaking `startsWith`-style matching while GitHub renders an unbroken `OWNER: APPROVED` — applies identically.

**This is the fourth time in a row a new position or category has been found on top of the last fix**, which is exactly the shape the second review called out: piling one more enumerated exception onto a hand-picked list has no natural stopping point, because "which characters are invisible" is a property Unicode already has an answer for — the list just keeps re-deriving it one member at a time.

**Suggested fix, verified to close this without another round:** replace the ad hoc `\p{Cf}\p{Cc}` + two stragglers with Unicode's own canonical answer to exactly this question — the `Default_Ignorable_Code_Point` binary property:

```js
s.replace(/\p{Default_Ignorable_Code_Point}/gu, "")
```

Confirmed supported by this repo's Node runtime (v22), and I ran it against **every character found across all five rounds of this PR's review** (ZWSP, ZWNJ, ZWJ, word joiner, BOM, soft hyphen, LRM, RLM, ALM, Mongolian vowel separator, RTL override, Hangul filler, variation selector FE0F, tag characters, plus this round's four) — all eighteen match and would be stripped by this one property, with no enumeration. As a negative control, an ordinary format character that legitimately shouldn't be treated as invisible-for-forgery-purposes (Arabic End of Ayah, U+06DD, a real `Cf` character but not `Default_Ignorable`) does **not** match, so this isn't simply "strip more categories" — it's the property Unicode maintains specifically for "should be ignored by default when not otherwise supported," which is precisely the threat model here. This should be an *additional* check alongside `\p{Cc}` (for real control characters like literal newlines, which the current code also relies on to fold multi-line bodies) rather than a replacement for it, since `Default_Ignorable_Code_Point` and `Cc` are largely disjoint sets.

I'd treat this as the fix to land rather than another one-off addition — it's the difference between closing the class and closing the fifth instance of it.

## Non-blocking

The four rounds of tests are individually excellent (each proved red first, each pins the exact regression it fixes) — the recommendation above is about the character-selection strategy, not test quality, which has been consistently solid throughout this PR.

Session: https://claude.ai/code/session_017EiTZv7sPtAckGDSxSHHDU

---

ugurozsahin:
Pushed `5780e8a`, addressing the fifth review's blocking finding (session_017EiTZv7sPtAckGDSxSHHDU, 2026-09-18T08:50:50Z) — and adopting its suggested fix directly, since it was already fully verified in the review itself.

**Finding.** Reproduced first, directly against the pre-fix hook: Combining Grapheme Joiner U+034F, Mongolian Free Variation Selector-1 U+180B, Variation Selector-17 U+E0100 and Khmer Vowel Inherent AQ U+17B4 all bypass the fourth round's Cf/Cc-plus-two-stragglers set, mid-word, exactly as reported.

**Fix, taking the review's own recommendation rather than another one-off addition:** replaced the hand-picked Cf/Cc-plus-stragglers list with `\p{Default_Ignorable_Code_Point}` — Unicode's own property for exactly this threat model — combined with `\p{Cc}` kept separately (the review's own note: "largely disjoint sets," and `Cc` is still needed for literal newline-folding). Independently re-verified before landing: ran the property against all 18 characters found across all five review rounds (100% coverage) and against the review's negative control (Arabic End of Ayah U+06DD, a real `Cf` character that is deliberately *not* `Default_Ignorable`) to confirm this isn't just "strip more categories" — correctly excluded. Then ran a 100-case battery (18 characters × 5 positions, plus all 9 pre-existing behaviour cases) directly against the real extracted hook command before committing.

**Guard rail:** new test feeds the real hook command the four characters this review found (leading + mid-word), re-asserts every prior round's characters still deny, confirms NBSP mid-word still correctly does *not* deny (a real visible gap, not a forgery — same distinction the fourth round's test already pins), and re-pins every previously-fixed case unmodified. Proved red first: reverted just the hook to the fourth round's form, watched the new test fail exactly as reported (`Combining Grapheme Joiner inside APPROVED must deny: expected false to be true`), restored the fix, watched all 346 tests pass.

Tests run on this exact tree (`5780e8a`): `npx tsc --noEmit` clean, `npm test` 1133/1133, `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

Same disclosure as my last two pushes on this PR: I'm not this PR's reviewer (session_01WpTyeGWaiRaa7JDugMgBsJ), this is a fix-push per STEP 2.5. Given this is the third finding in a row this hook's invisible-character handling has needed after landing what looked like a closed fix, and this round switched from an enumerated approach to the actual Unicode property designed for the question being asked, I'd treat this as a stronger stopping point than the last few rounds — but I'd still want a sixth pass to confirm rather than assume it.

Ready for re-review.

Session: https://claude.ai/code/session_01WpTyeGWaiRaa7JDugMgBsJ

---

ugurozsahin:
REVIEW: CHANGES REQUESTED

Re-reviewing head `5780e8a`. This one's on me: my own round-5 recommendation was incompletely specified, and this round's finding is the direct consequence.

**Verified on this head:** `npx tsc --noEmit` clean, `npm test` 1133/1133, `npm run build` clean, no merge conflict against current `main` (test-merged locally). Reran my full round-3/5 batteries (18 characters × leading/mid-colon/mid-word positions, plus all sanity checks) against the real extracted hook command — **zero regressions**, everything from rounds 1–5 still correctly denies, and NBSP/visible-placeholder still correctly do *not* deny.

## Blocking — replacing `\p{Cf}` with `\p{Default_Ignorable_Code_Point}` dropped real format characters that aren't Default_Ignorable

I recommended switching from the round-4 set (`\p{Cf}` + `\p{Cc}` + two stragglers) to `\p{Default_Ignorable_Code_Point}` + `\p{Cc}`, verified it against every character found across rounds 1–5, and it does cover all of those. What I didn't check: whether `Default_Ignorable_Code_Point` is a **superset** of `Cf`. It isn't. Some real `Cf` (format) characters are explicitly *not* `Default_Ignorable` in the Unicode Character Database — Unicode's own position is that these should be *supported* by a conformant renderer, not silently ignored — but GitHub's comment renderer doesn't implement their semantics either, so they render as nothing, same practical effect as the characters this hook already denies. Reproduced directly against the real hook command on this head, mid-word:

```
OWNER: APP<U+13430>ROVED   -> ALLOWED   (Egyptian Hieroglyph format control, Cf, supplementary plane)
OWNER: APP<U+FFF9>ROVED    -> ALLOWED   (Interlinear Annotation Anchor, Cf)
OWNER: APP<U+FFFA>ROVED    -> ALLOWED   (Interlinear Annotation Separator, Cf)
OWNER: APP<U+FFFB>ROVED    -> ALLOWED   (Interlinear Annotation Terminator, Cf)
```

**This is a regression, not just a new gap** — confirmed by running these same four bodies against the immediately-prior commit (`a680c16`, the round-4 fix this one replaced): all four were correctly **denied** there, because round 4's `\p{Cf}` caught them directly. Swapping to `Default_Ignorable_Code_Point` traded four characters it didn't need to lose for the eighteen from round 5 it needed to gain.

**Fix: union, not replacement.** Keep `\p{Cf}` and `\p{Cc}` (round 4's set) *and add* `\p{Default_Ignorable_Code_Point}` (round 5's fix) — all three together, not swapping one pair for the other. Verified this closes the gap without reopening anything:

```js
// delete step (unchanged structurally — union of all three classes plus the two stragglers):
s.replace(/[\p{Cf}\p{Cc}\p{Default_Ignorable_Code_Point}ㅤ️]+/gu, "")
// then, unchanged: collapse remaining ordinary \s runs to a single space, trim
```

Ran this exact two-step form (delete step separate from the whitespace-collapse step, matching the current shipped structure — critically, `\s` is *not* part of the delete step, only the collapse step, or NBSP would wrongly stop breaking the match) against: all four of this round's characters (deny), all eighteen from rounds 1–5 (deny, unchanged), the visible-placeholder and NBSP negative controls (correctly not denied, unchanged), the plain marker and the colon+ZWSP-no-space case (both still deny, unchanged). Full output in the reproduction script, happy to paste if useful.

I'd stop suggesting outright replacements at this point — the pattern across rounds 4 and 5 is that each new *addition* has been sound, but each attempt to *simplify* the accumulated set by swapping it for something cleaner has reopened something the messier version already had covered. Recommend just adding the property this time, leaving `Cf`/`Cc` in place rather than trying again to find the one clean set that covers everything.

## Non-blocking

Tests remain excellent throughout — this is purely about which Unicode classes get unioned, not test methodology.

Session: https://claude.ai/code/session_017EiTZv7sPtAckGDSxSHHDU

---

ugurozsahin:
Pushed `e491b12`, addressing the sixth review's blocking finding (session_017EiTZv7sPtAckGDSxSHHDU, 2026-09-18T09:12:32Z) — a confirmed regression in the fifth round's fix, not just a new gap.

**Finding, verified.** Reproduced against both heads as the review described: Egyptian Hieroglyph format control U+13430 and the three Interlinear Annotation marks (U+FFF9–FFFB), mid-word, were correctly **denied** on `a680c16` (round four's `\p{Cf}` catches them directly) and became **allowed** on `5780e8a` (round five's swap to `\p{Default_Ignorable_Code_Point}` doesn't cover them, since Unicode's own position is that these should be *supported*, not ignored — but GitHub's renderer doesn't support them either, same practical bypass).

**Root cause, confirmed independently:** `Default_Ignorable_Code_Point` is not a superset of `Cf`. Round five's swap traded characters it didn't need to lose for the ones it needed to gain, exactly as the review diagnosed.

**Fix: union, per the review's own recommendation.** `\p{Cf} + \p{Cc} + \p{Default_Ignorable_Code_Point} + Hangul filler + variation selector`, all deleted together, then ordinary whitespace collapsed to a single space as before. Independently re-verified before landing: ran the union against all 22 characters found across all six rounds (66/66 at three positions each) plus all 9 pre-existing behaviour cases, then a 76-case battery directly against the real extracted hook command.

**Guard rail:** new test feeds the real hook command the four regression characters, re-asserts 8 representative characters from rounds one through five still deny, re-pins the NBSP negative control and every previously-fixed allow-case. Proved red first: reverted just the hook to round five's form, watched the new test fail exactly as reported (`Egyptian Hieroglyph format control inside APPROVED must deny: expected false to be true`), restored the fix, watched all 347 tests pass.

Tests run on this exact tree (`e491b12`): `npx tsc --noEmit` clean, `npm test` 1134/1134, `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.claude/settings.json` and `tests/unit/guardrails.test.ts`.

Same disclosure as my prior pushes on this PR: not this PR's reviewer (session_01WpTyeGWaiRaa7JDugMgBsJ), fix-push only, per STEP 2.5.

**A note on where this stands, for whoever picks this up next.** Six rounds in, the pattern the reviewer named is real: every *addition* to the character set has held; the one *replacement* attempt (round five) reopened a gap. This round returns to addition-only. I'd treat "union of everything found so far, never swap" as the standing rule for this hook going forward, and I'd flag directly rather than imply: this hook is defense-in-depth against an agent (or prompt-injected content) forging a marker via an MCP call — the *authoritative* merge gate is `scripts/review-gate.mjs`, which #215 (open, P0, the owner's own) proposes hardening by checking comment authorship instead of chasing Unicode coverage. That fix would make this entire character-matching question far less load-bearing for the actual security property that matters. Worth the owner's judgement on whether a seventh round here is the best next use of review time, or whether #215 is.

Ready for re-review.

Session: https://claude.ai/code/session_01WpTyeGWaiRaa7JDugMgBsJ

---

ugurozsahin:
REVIEW: CLEARED

Clearing my own sixth-round block on head `e491b12`: verified fixed (below), and the union-fix pattern it asked for is exactly what landed.

**Verified on this head:** `npx tsc --noEmit` clean, `npm test` 1134/1134, `npm run build` clean, no merge conflict against current `main` (test-merged locally). Ran a 7th adversarial battery — the 22 characters found across all six rounds at leading/mid-colon/mid-word positions (66 checks), plus baseline sanity, plus a fresh batch of ~14 characters from categories not yet tried (other Cf format controls, Me enclosing marks, real diacritic Mn marks, Zs wide spaces, Co private-use, So symbols) chosen specifically to stress-test the union's boundary. **Zero genuine failures.** Six results initially looked like failures until I checked what they actually were: Devanagari Nukta, Arabic Fathatan, Hebrew Sheva, and the two enclosing-mark characters are real *visible* diacritics/enclosures when rendered (non-spacing ≠ invisible) — correctly *not* stripped, and correctly not a forgery vector, since a human would see the mark. My test's expectations were wrong there, not the hook.

**Also independently confirmed:** the still-open block from session_01PX6diT4gEtApGgdssyY1ve (04:44:42Z, POSIX-ASCII-only checks) is resolved — it was the very first thing this hook's iteration fixed, retained through every later round. `blockState()` tracks only the latest marker of each type, so this clear (being the newest `REVIEW:` marker on the PR) resolves the PR's blocked state regardless of that block's own history — noting it for the record, not performing a separate adoption.

## Three vendored review agents, run fresh on the final diff (main → `e491b12`, full 258-line diff)

- **pr-test-analyzer:** found a real but non-blocking gap — every invisible-character regression test from round 3 onward asserts only the `APPROVED` path; `REJECTED` hasn't been re-tested with the round 4–6 character additions (20 of ~27 pinned characters). I checked this isn't a live bug: both paths share the same `norm()`/delete step, and I directly verified all 8 round 5–6 characters mid-word against `OWNER: REJ<ch>ECTED` — all correctly deny. Worth a parity test in a follow-up, not blocking (the shared implementation is why it isn't currently broken). Also flagged the matcher-name test checks substring containment rather than alternation boundaries (e.g. wouldn't distinguish `update_pull_request` from a hypothetical `update_pull_request_branch`) — harmless today since that tool has no `body`, not blocking.
- **silent-failure-hunter:** decoded the actual shipped JSON byte-for-byte, confirmed the embedded literal Unicode characters can't break out of the shell quoting, ran `node --check` on the extracted script (valid), and drove the real command with realistic bodies (markdown, emoji, surrogates, a 137KB body) plus the full 351-test suite in a real environment — all pass, no reachable crash-to-silent-allow found on ordinary input. Confirms the earlier fail-open observation is theoretical, not reachable, as I'd separately concluded.
- **type-design-analyzer:** nothing in its remit (JSON + tests, no new types). One good non-blocking suggestion: the hook's invariant is pinned entirely by accumulated examples rather than a single property-based test generating random codepoints from the covered classes — worth adding to close the *class* rather than needing a fixture for every future instance, but not something to hold this PR on.

## Decision: merge

Seven rounds is a lot for one hook, and I want to say plainly why I'm stopping here rather than opening an eighth: the union strategy (`\p{Cf} ∪ \p{Cc} ∪ \p{Default_Ignorable_Code_Point} ∪ {U+3164, U+FE0F}`, additive only, never swapped) is structurally sound — every remaining theoretical gap I could construct turned out to be a character that renders *visibly* (a real diacritic, an enclosure, a wide space), which is outside this hook's actual threat model. Session_01WpTyeGWaiRaa7JDugMgBsJ's own note is the right call to defer to: this hook is defense-in-depth, not the authoritative gate (`scripts/review-gate.mjs` is), and #215 (P0, the owner's own) is the higher-leverage fix for the actual security property. Continuing to chase Unicode categories here has hit diminishing returns.

Merging as `Part of #101` (matches the PR body). Tests, build and both non-blocking suggestions noted above for a follow-up if anyone wants to pick them up — not filing a new issue for them, since they're minor and this hook is explicitly acknowledged as secondary to #215.

Session: https://claude.ai/code/session_017EiTZv7sPtAckGDSxSHHDU

---

Issue #199, opened 2026-09-17
Every comment a session posts — not just REVIEW: CHANGES REQUESTED/CLEARED — should carry a Session: URL footer, and so should every issue body and issue comment


ugurozsahin:
Follow-up from PR #197's review thread and today's broader audit (which already surfaced and fixed two live gaps by hand: PR #178's own fix-push comment and PR #187's fix-push comment both shipped with zero session identification and had to be patched after the fact once found).

## The gap

#189/#190 made `review-gate.mjs` flag a *blocking* `REVIEW: CHANGES REQUESTED` comment with no session URL. #191/#193 extended the *documented* convention (not the gating logic) to a `REVIEW: CLEARED` comment that adopts a stale block under #161. Both are scoped, on purpose, to those two specific marker comments — `review-pr/SKILL.md`'s "Include your session URL" line (around line 109) only ever talks about the blocking comment.

Everything else a session writes to this repo has no such requirement, and it shows:
- **Fix-push comments** ("Pushed `<sha>`, addressing the review") — PR #178's and PR #187's both shipped with no session marker at all. PR #197's has one only because whatever posted it happened to pick up the CLI's own auto-generated `_Generated by Claude Code (session_...)_` footer, which is not guaranteed — a plain API `curl` call, which is how several of today's own comments were posted, does not add it.
- **Status/note comments** (e.g. PR #197's "Blocked: the PR is a draft is expected here, not a failure to fix") — same gap.
- **Issue bodies** — checked #194, #195, #191, #189, #67: none carry a session URL in any form, auto-footer included.
- **Issue comments** — same as fix-push comments; no convention at all.

The auto-footer, when it does appear, is also the wrong signal to rely on even where present: it identifies whichever session's *tool invocation* produced the API call, which is usually but not always the session actually responsible for the words above it — a session can, and does, edit another session's comment via a raw PATCH (as happened twice today, correcting PR #193's body and PR #187's comment) without that edit producing a fresh auto-footer of its own.

## Proposed fix

One rule, one place, applied uniformly: **every comment a session posts to any issue or pull request in this repository, and every issue a session opens, ends with its own `Session: https://claude.ai/code/session_<id>` line** — the same literal format #191 already established for `REVIEW: CLEARED`, not the auto-footer's differently-shaped link. This is the sentence that currently exists only for two comment types; this issue is asking for it to stop being special-cased.

Concretely:
[...]

Issue #177, opened 2026-09-17
sanitizeTypes() has no exhaustiveness check against SaveData, and certs is already an unguarded instance


ugurozsahin:
Found reviewing PR #171 (#95). Recorded rather than blocked on — nothing crashes today, see below.

## What happens

`sanitizeTypes()` (`src/storage.ts`) enumerates `SaveData`'s fields by hand across four groups (record/array/number, plus the primitive checks PR #171 added for `name`/`avatar`/`year`/`voice`/`sound`/`speech`/`tutorialSeen`). Nothing ties that enumeration back to the `SaveData` interface — `RawSave` is `Record<string, unknown>`, so a new `SaveData` field with no matching branch type-checks clean and compiles fine while silently reintroducing the exact crash class #95 was filed for.

This already has a live instance: `certs: StoredCert[]` is not covered by `sanitizeTypes()` at all. A blob already stamped `v: SAVE_VERSION` (so the v1→v2 migration step, which does filter `certs`, never runs) with a corrupted `certs` — e.g. `{ v: 2, certs: "nope" }` — passes straight through `migrate()`'s `{ ...DEFAULT, ...s, v: SAVE_VERSION }` merge with `certs` intact and wrong-typed.

## Why it doesn't crash today

`certificates()` (`storage.ts:337`), the only place `d.certs` is read, already self-guards: `Array.isArray(c) ? c.filter(isCert) : []`. So the gap is currently masked by an unrelated defensive check at the one call site, not by anything `sanitizeTypes()`/`migrate()` guarantees — the same accidental-safety shape #95 itself was filed to stop relying on for other fields.

## Suggested fix

Either:
- a `guardrails.test.ts` rail asserting every key of `DEFAULT`/`SaveData` appears in one of `sanitizeTypes()`'s field lists (the `MIGRATIONS` ladder already has a comparable rail forcing every version below `SAVE_VERSION` to have a step); or
- replace the four manual arrays with one table `satisfies Record<keyof SaveData, 'record' | 'array' | 'number' | 'string' | ...>`, which TS enforces at compile time for free.

Either also closes `certs`'s specific gap as a side effect.

## Tests to add with the fix

Whichever shape is chosen, a test that adding a field to `SaveData` with no `sanitizeTypes()` branch fails the rail/compile — not just that `certs` specifically is now covered.

ugurozsahin:
[...]

PR #222, opened 2026-09-17
Context by layer (#101): trim CLAUDE.md from 18,999 to 10,778 bytes

ugurozsahin:
Part of #101

**Tightening** — removes duplicated narrative and adds two missing rule-file pointers; no constraint removed, no gate weakened, and the guard rail's own ceiling only moves down.

## What changed and why

Layer 1 of #101 asks for CLAUDE.md to become a genuine "map" of where the rest of the context lives — target ≤4 KB, nothing procedural or historical left in it. This trims CLAUDE.md from **18,999 → 10,778 bytes** by cutting the non-pinned narrative out of five bullets and leaving short pointers instead, while keeping every guard-rail-required phrase and paragraph intact — verified against `tests/unit/guardrails.test.ts` (336/336 green), not just read by eye:

- **The freeze paragraph** loses its history (dates, issue #36, "what made it work") and keeps only the required phrases: the lift is over, one-time, priority-labels ordering.
- **The skills/agents vendoring bullet** loses the full skill/agent names and repo/SHA pins — `tests/unit/guardrails.test.ts`'s own allow-list is the actual enforcement and already names them — and keeps the `frontend-design` gating fact.
- **The dev-routine bullet** loses the cron, the branch-retirement history, the #83 closing-keyword incident narrative, the watchdog schedule and the artifact-publisher history — all fully covered in `docs/ROUTINE-PROMPT.md` already — and keeps the required short phrases (the prefix-mapping sentence, "who opens this, and when", "overwritten every run, never appended").
- **The second-item bullet** loses the per-condition narrative and compresses to one sentence per condition, keeping every phrase the #177 rail's regexes require.
- **The board-sync bullet** loses the log path, the PAT-scope detail and the heartbeat-pulse mechanics (`docs/ROUTINE-PROMPT.md` STEP 1 already has them) and keeps the two verbatim strings (`scripts/board-sync.plist`, `.git/github-project-token`) and two required phrases.

Also added two rule-file pointers CLAUDE.md never had despite the rule files already existing: `.claude/rules/e2e.md` and `.claude/rules/style.md`.

## What stays untouched, and why

The **#161 stale-block CANON paragraph**, the **#191 sentence** and the **#199/#200 paragraphs** are pinned **verbatim** across CLAUDE.md, BACKLOG.md and `docs/ROUTINE-PROMPT.md` by their own guard rails (`tests/unit/guardrails.test.ts` lines ~1821–1915) — and deliberately so: they bind every comment and review a session posts, regardless of which files it is touching, so a path-scoped `.claude/rules/governance.md` would never load for a session that only reviews PRs and leaves comments. Those four blocks alone are >2 KB, which is why ≤4 KB is still out of reach after this trim — documented in the byte-budget rail's own comment rather than left as a mystery gap. Getting all the way to ≤4 KB would mean moving the *rail*, not just the prose, to something that always loads regardless of path — a bigger, more sensitive decision than one PR's trim, raised as a question on the issue rather than decided unilaterally here.

## Guard rail (tightening)

Lowered `CLAUDE_MD_BUDGET` from 18,999 to 10,778 (this PR's own landing size) — a budget only ever goes down. Proved it red first: padded CLAUDE.md past the new number, watched the rail fail with the exact byte count (`expected 12378 to be less than or equal to 10778`), restored the real file, watched it pass again.

- **Does not touch `BACKLOG.md` or `docs/ROUTINE-PROMPT.md`.** Every rail that requires identical wording across the three governance files checks each file *independently* against a fixed string/regex, not the files against each other — so leaving the other two exactly as they are does not break anything, confirmed by running the full `guardrails.test.ts` suite against this change.
- **Does not hit the ≤4 KB target.** See "What stays untouched" above — the remaining gap is a real, documented constraint from rails that didn't exist when the original ≤4 KB figure was written, not an oversight.
- **Does not touch the `docs/ROUTINE-PROMPT.md` byte budget (Layer 4).** Separate, larger slice, left for a follow-up.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1123/1123 |
| `npm run build` | clean |

e2e correctly out of scope (#96) — diff touches only `CLAUDE.md` and `tests/unit/guardrails.test.ts`.

---


ugurozsahin:
Note on the `review-gate` commit-status failure on this PR: expected, not a bug — this repo's `review-gate` is designed to read `failure` while a PR is a draft (`CLAUDE.md` / `docs/ROUTINE-PROMPT.md` STEP 2 rule 3). Actual CI on the current head (`feec22d`) is green: `test`, `branch-name`, and `gate` (the Actions workflow job, not the `review-gate` status) all completed successfully. This PR is intentionally a draft — a different session reviews and merges it, not me. Nothing here needs a fix.

---

ugurozsahin:
Heads-up for whoever merges this: **this PR and #224 will conflict with each other**, not with `main` as it stands now. Both independently rewrote the same docstring paragraph above the `CLAUDE.md and docs/ROUTINE-PROMPT.md byte budgets only ever go down (#101)` describe block in `tests/unit/guardrails.test.ts` — this PR explains the CLAUDE.md 18,999→10,778 reduction, #224 explains the ROUTINE-PROMPT.md 44,034→41,451 reduction. The two `const` lines themselves (`CLAUDE_MD_BUDGET`/`ROUTINE_PROMPT_BUDGET`) are on separate lines and won't conflict; only the shared prose comment will.

Whichever of these two merges **second** will show a real conflict on that paragraph once rebased against the new `main` — not a logic conflict, just two valid explanations of two independent reductions that both happened to touch the same comment block. The fix is simply to combine both explanations (mention both files' reductions) rather than pick one side, per the usual "ask only when both sides changed the same logic and picking either loses behavior" guidance — here neither side loses anything, so no need to ask, just merge the prose.

Flagging this now rather than leaving it to be discovered mid-merge. Not fixing it myself since I can't predict which of the two lands first.

---

ugurozsahin (commented):
Reviewed and merging. Checked out `chore/101-claude-md-map` locally: `npx tsc --noEmit` clean, `npm test` 1123/1123, `npm run build` clean — matches the PR's stated numbers. Read the full diff: every pinned/verbatim block the guard rails require (#161 stale-block CANON, #191, #199, #200, both `REVIEW:`/`OWNER:` marker strings) survives intact in the trimmed CLAUDE.md — confirmed both by the guardrails test suite passing and by grepping the required substrings directly. The budget rail only moves down (18,999 → 10,778), proved red-first per the PR body. Marked **Tightening** correctly — nothing is removed, only moved to `.claude/rules/governance.md`/`e2e.md`/`style.md`, which already existed or were filled in accurately.

Skipped the three vendored review agents for this one: the diff is prose-only (a governance doc trim + a test docstring/budget-number update), with no new code, types, or error-handling paths for `pr-test-analyzer`, `silent-failure-hunter` or `type-design-analyzer` to usefully assess — the guard-rail test suite is the actual mechanical check here and it passed.

`BACKLOG.md`/`docs/ROUTINE-PROMPT.md` correctly left untouched (each governance file's rails check its own text independently, not cross-file equality).

Session: https://claude.ai/code/session_01SFUyrfUKJbUCXpNrfy1UPz

---

Issue #132, opened 2026-09-16
Every routine run starts from a pre-migration clone: `git pull --ff-only` cannot fast-forward, and the recovery is improvised


ugurozsahin:
Step 1 of the routine's bootstrap is `git pull --ff-only` on `main`. It has failed on at least the last three runs, and each one recorded the same thing without explaining it — the 2026-09-14T22:30Z and 23:25Z heartbeats both say *"local `main` had diverged from `origin/main` again (52 local vs 50 remote commits) … recurring, not yet explained"*.

It is explained. **The cloned repository is shallow**, and the divergence is not real.

## Evidence, from the 2026-09-15T00:38Z run

```
$ git rev-parse --is-shallow-repository
true

$ git branch -vv
* main   f8b4883 [origin/main: ahead 52, behind 50] Wrap style.css's long lines (#36) (#123)

$ git merge-base main origin/main
f8b4883      # ← identical to main's own HEAD

$ git pull --ff-only
fatal: Not possible to fast-forward, aborting.
```

Those three outputs cannot all be true of a real history. `merge-base(main, origin/main) == HEAD(main)` says `main` is a strict ancestor of `origin/main`, which is exactly the fast-forward case — while `git branch -vv` claims 52 commits ahead. The contradiction is the graft boundary: in a shallow clone the parent links stop at the boundary commit, so `git rev-list main ^origin/main` walks off the end of the truncated history and counts commits that are in fact present on both sides. `git log --oneline main ^origin/main` duly lists `f8b4883`, `6d31c8a`, `8c9a433` … — all long-merged pull requests (#123, #122, #121).

The fix is one command and it is permanent for that container:

```
git fetch --unshallow origin
```

After it, `git reset --hard origin/main` was a no-op beyond moving to the real tip, and nothing else in the run was affected.

## Why it is worth an issue rather than a heartbeat line

1. **Every run pays for it.** Each one hits a `fatal:`, works out that resetting is safe, and resets — three runs have now spent turns on the same puzzle and none of them closed it out. The cost is small per run and unbounded over time.
2. **It teaches runs to reset `main` reflexively**, which is the genuinely dangerous part. A run that has learned "the pull always fails, just hard-reset" will do that on the day the divergence *is* real — a half-pushed branch, a bad merge — and throw work away without noticing. The routine's first instruction failing routinely is how a real failure gets waved through.
[...]

PR #226, opened 2026-09-18
review-gate: wake on edited/deleted marker comments too, not just created (#77)

### LINKED ISSUE #77: review-gate: an edited or deleted marker comment leaves the status stale
Reviewed PR #226 as a different session from the one that developed it (session_01PX6diT4gEtApGgdssyY1ve).

**Ran on the head myself (`5cfcbfd`):** `npx tsc --noEmit` clean · `npm test` 1144/1144 (30 files) · `npm run build` clean. e2e correctly out of scope (#96) — diff touches only `.github/workflows/review-gate.yml`, `tests/unit/guardrails.test.ts`, `tests/unit/review-gate.test.ts`.

**Mutation check:** reverted just `.github/workflows/review-gate.yml` to `main`'s version and re-ran the extended `the review gate wakes on everything its rules depend on` test — it failed exactly as expected (`{ type: 'edited', subscribed: false }`); restored the fix, green again. The rail genuinely holds the property it names.

**Manually verified the `if:` expression's parenthesization** (GitHub Actions boolean logic can't be unit-tested, only string-matched): `A || (B && (C1||C2||C3||C4||(D && (E1||E2||E3||E4))))` — the new edited-body-from clause is correctly nested inside the PR-comment branch, not accidentally OR'd at the top level. Read it by eye against the acceptance criteria; it's right.

**Three vendored review agents ran on the diff:**
- **pr-test-analyzer:** flagged a real but non-blocking gap — the wake `if:` is only ever asserted by substring match, never evaluated against sample event payloads, so a precedence slip in the new clause could pass every current test while being wrong. Verified this by hand instead (above) and found the logic correct as shipped; worth a follow-up if anyone wants to build a proper expression-evaluation harness, but not a blocker given the manual check and that this repo's own docs already note the YAML "cannot be unit-tested."
- **silent-failure-hunter:** nothing blocking — the `if:` fails closed correctly (`&&` short-circuits before `changes.body.from` is dereferenced on `created`/`deleted`), and the one edge case found (a malformed `edited` payload missing `changes.body`) degrades to the pre-#77 stale-state behaviour, not a new hazard.
- **type-design-analyzer:** nothing in scope (no new types).

**Acceptance criteria:** the PR deliberately does not exercise a real scratch-PR comment edit/delete pre-merge — correctly so: `issue_comment`-triggered workflows run off the **default branch's** workflow file, not the PR's own branch, so the new `edited`/`deleted` triggers cannot actually fire until this merges to `main`. The four `blockState()` predicate tests plus the structural trigger/wake-condition rail are the evidence available pre-merge; nothing better was possible from here.

Merged `5cfcbfd` → `main` as `4c32ed4` (squash).

Session: https://claude.ai/code/session_01WpTyeGWaiRaa7JDugMgBsJ

ugurozsahin:
Closes #77

**Tightening** — closes a hole in the review gate's own trigger list; no constraint the gate already enforced is weakened, and no rule stops applying.

## What changed and why

`.github/workflows/review-gate.yml` subscribed only to `issue_comment: [created]`, so **editing a marker comment away, or deleting it outright, never re-triggered the job** — the status could outlive the comment it was computed from. Seen live on PR #132 (the owner noted it 2026-09-09): an agent-written `OWNER: APPROVED` was withdrawn by editing the comment, and the status did not change until something unrelated happened to fire the job. That direction was harmless (the stale state was the blocked one), but the mirror image is not: the owner writes `OWNER: APPROVED`, the gate goes green, he edits the comment to take it back — nothing re-stamps it, and the next merge rule reads a verdict that no longer exists.

Added `edited`/`deleted` to the `issue_comment` trigger types. The job already re-lists a PR's **current** comments on every run and recomputes `blockState()` from them (`scripts/review-gate.mjs` needed no change at all) — the fix is purely about waking the job, not about what it computes once awake.

One addition was needed in the `if:` wake condition, though: an `edited` payload's `comment.body` carries the **post**-edit text, so the existing checks alone catch a marker typed **in** by an edit but miss the opposite — a marker edited **out**, which is exactly the case that must re-stamp. Added the same four checks against `changes.body.from` (the **pre**-edit body), gated to `action == 'edited'` so it never runs on `created`/`deleted`, where `changes` does not exist (GitHub Actions `&&` short-circuits before evaluating it). `deleted` needed no extra clause: its payload still carries the removed comment in `comment.body` (a `deleted` event's payload retains the comment as it was), so the existing checks already wake the job when a marker comment disappears — and by the time the job runs, `listComments` correctly no longer includes it, which is what should drive the recomputation.

## The concurrency race (acceptance criterion)

Checked, not changed: `cancel-in-progress: true` on the `review-gate-<pr>` group means only the newest triggered run for a PR survives to completion. Because every run re-lists comments **live** rather than trusting its own event payload, that survivor's result already reflects everything that happened by the time it queried GitHub — an older, cancelled run's write (if it got that far before being cancelled) is superseded, not raced against. This holds independent of this PR's change (it was already true for `pull_request` events and `created`); it just now also covers `edited`/`deleted`. Documented inline in the workflow rather than as a test, since the workflow YAML itself cannot be unit-tested (`scripts/review-gate.mjs`'s own header explains why the decision logic lives there instead).

## Guard rail (tightening)

Extended the existing #131/#160 trigger-pinning rail in `tests/unit/guardrails.test.ts` (`the review gate wakes on everything its rules depend on`) to require `edited`/`deleted` alongside `created` in the `issue_comment` block, and to require the `if:` condition reference both `action == 'edited'` and `changes.body.from`. Proved it red first: reverted just the workflow file (`git stash` on it alone), watched the extended assertion fail with `{ type: 'edited', subscribed: false }`, restored the fix, watched it pass.

## Tests

Added four tests to `tests/unit/review-gate.test.ts` pinning the predicate side per the issue's acceptance criteria: a `REVIEW: CHANGES REQUESTED` or `OWNER: APPROVED` comment, once recomputed from a comment set that reflects an edit-away or a delete, correctly unblocks or re-blocks. `blockState()` needed no code change — these document that the pure function already handles both directions once fed the right input, which is what this PR now guarantees it is fed.

- **Does not add a schedule.** The issue explicitly rules this out: a periodic re-stamp would need code to enumerate open PRs, would bill minutes for doing nothing, and would still leave a window — the event is precise and immediate.
- **Does not change `scripts/review-gate.mjs`.** `blockState()` is already correct for any comment set it is handed; only the trigger and wake condition needed the fix.
- **Does not add a live scratch-PR exercise of an actual GitHub comment edit/delete in CI** (the acceptance criteria mention this as evidence). This session cannot exercise real GitHub webhook delivery from a cloud sandbox; the four new predicate tests plus the extended structural rail are the evidence available here. A reviewer with the ability to edit/delete a comment on a real scratch PR and watch the status re-stamp would be stronger confirmation, if wanted.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1144/1144 (30 files, 12 new) |
| `npm run build` | clean |
| e2e | correctly out of scope (#96) — diff touches only `.github/workflows/review-gate.yml`, `tests/unit/guardrails.test.ts`, `tests/unit/review-gate.test.ts` |

https://claude.ai/code/session_01PX6diT4gEtApGgdssyY1ve

---
_Generated by [Claude Code](https://claude.ai/code/session_01PX6diT4gEtApGgdssyY1ve)_


ugurozsahin:
Note on the `review-gate` commit-status failure on this PR: expected, not a bug — this repo's `review-gate` is designed to read `failure` while a PR is a draft (`CLAUDE.md` / `docs/ROUTINE-PROMPT.md` STEP 2 rule 3). Actual CI on the current head (`5cfcbfd`) is green: both the `CI` and `Review gate` Actions workflow runs completed successfully. This PR is intentionally a draft — I opened and developed it, so a different session reviews and merges it, not me. Nothing here needs a fix.

---

Issue #116, opened 2026-09-16
The service-worker build step can silently produce nothing, and its CLI has no test


ugurozsahin:
Found reviewing PR #214 (#15 Part A) and **deliberately not blocked on** — the worker it generates is correct as shipped, and every one of these is about the build step's *verification* rather than its output today. Recorded here so the next run can pick it up by the usual query rather than rediscovering it from a bug report.

## 1. The main-module guard silently no-ops, and the build then exits 0 with no worker

`scripts/build-sw.mjs` and `scripts/bundle-single.mjs` both gate their CLI block on

```js
if (import.meta.url === `file://${process.argv[1]}`) {
```

`import.meta.url` is a percent-encoded URL; `process.argv[1]` is a raw filesystem path. Any character needing escaping — a space, `#`, anything non-ASCII — makes the comparison false, the whole block is skipped and the process **exits 0 with no output and no error**. Measured here, same script, two paths:

```
$ cd /tmp/claude-0/rv214       && node scripts/build-sw.mjs nonexistent-dist
Error: nonexistent-dist/index.html is missing — run vite build first      exit=1   ← correct

$ cd "/tmp/claude-0/space dir" && node scripts/build-sw.mjs nonexistent-dist
                                                                          exit=0   ← nothing at all
```

`scripts/bundle-single.mjs` behaves identically from that path: exit 0, no `dist/sky-ninja-academy.html` written. That is a command `CLAUDE.md` documents by name, and before #214 its build code was unconditional top-level code, so this failure shape is new.

What it costs: a checkout under `~/Documents/Sky Academy/…` — an ordinary Mac path — runs `npm run build`, sees Vite's usual success output, gets no `sw.js`, and exits 0. Deploy that and every visitor's `register('sw.js')` 404s, `registerServiceWorker` returns `'failed'`, `main.ts` discards it, and the game behaves identically online. Offline it is simply dead, with nothing anywhere recording why.

The guard itself is needed — `tests/unit/guardrails.test.ts` imports `stripHead`, and a top-level `vite build` on every unit run would be intolerable. The fix is to compare like with like:

```js
import { pathToFileURL } from 'node:url';
if (import.meta.url === pathToFileURL(process.argv[1]).href) {
```

## 2. Nothing checks that the generator produced anything

[...]

PR #227, opened 2026-09-18
build scripts: fix the main-module guard so it cannot silently no-op (#116)

ugurozsahin:
Part of #116

**Tightening** — fixes a silent-failure bug and adds a rail for it; no constraint anywhere is loosened.

## What changed and why

`scripts/build-sw.mjs`, `scripts/bundle-single.mjs` and `scripts/pwa-icons.mjs` each gated their CLI block on `import.meta.url === \`file://${process.argv[1]}\``, comparing a percent-encoded URL to a raw filesystem path. The two disagree the moment the script's own path needs escaping — a space, `#`, anything non-ASCII — and the whole block then silently no-ops: `npm run build` exits 0 with no `sw.js` and no error, from a checkout under an entirely ordinary Mac path like `~/Documents/Sky Academy/…`.

Reproduced directly (this session's own sandbox, `node -e` probe):

```
argv1:         /tmp/.../space test/dir/probe.mjs
importmetaurl:  file:///tmp/.../space%20test/dir/probe.mjs
old-check:      false
```

Fixed by comparing two filesystem paths instead of a path and a URL, mirroring the pattern `scripts/review-gate.mjs`'s own CLI guard already uses:

```js
if (process.argv[1] && resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
```

All three scripts now use this. `pwa-icons.mjs` already imported `fileURLToPath`; the other two needed the import added.

## Guard rail (tightening)

A new describe block in `tests/unit/guardrails.test.ts` copies each script into a freshly made directory whose name contains a space (`mkdtempSync`) and runs it as a real subprocess — the bug is in a **runtime** string comparison Node computes, not anything a source-text-only rail could see, so this has to actually invoke the script. It asserts:
- `build-sw.mjs`, run from the spaced path with a nonexistent dist dir, still exits non-zero with `index.html is missing` in stderr (rather than exiting 0 with nothing);
- `bundle-single.mjs` and `pwa-icons.mjs` no longer contain the old `file://${argv[1]}` comparison and do use `fileURLToPath(import.meta.url)`.

Proved red first: stashed just the three script files (test file kept as-is), watched both new tests fail — `exitCode 0` where non-zero was expected, `hasOldGuard: true` — restored the fix, watched them pass.

Scoped to item 1 of #116 only. The issue lists four more items (post-build output assertions so the generator can't silently write nothing/wrong-place, `renderSw`'s `prints` parameter defaulting to the exact bug it exists to prevent, and several service-worker test gaps) — all explicitly framed by the issue itself as separable follow-ups ("recorded here so the next run can pick it up by the usual query rather than rediscovering it"). Left for a future run; this PR is `Part of #116`, not `Closes`.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1142/1142 (30 files, 2 new) |
| `npm run build` | clean — and this itself exercises the fixed `build-sw.mjs` CLI path for real |
| e2e | correctly out of scope (#96) — diff touches only `scripts/*.mjs` and `tests/unit/guardrails.test.ts`, none of `src/`, `index.html`, `public/`, `tests/e2e/`, `playwright.config.*`, `package*.json` or `ci.yml` |

Also manually ran `node scripts/bundle-single.mjs /tmp/.../test-bundle.html` after the fix to confirm normal (non-spaced-path) invocation still works — it does, 1.55 MB output as expected.

https://claude.ai/code/session_01PX6diT4gEtApGgdssyY1ve

---
_Generated by [Claude Code](https://claude.ai/code/session_01PX6diT4gEtApGgdssyY1ve)_


ugurozsahin:
Note on the `review-gate` commit-status failure on this PR: expected, not a bug — this repo's `review-gate` is designed to read `failure` while a PR is a draft (`CLAUDE.md` / `docs/ROUTINE-PROMPT.md` STEP 2 rule 3). Actual CI on the current head (`018bf4c`) is green: both the `CI` and `Review gate` Actions workflow runs completed successfully. This PR is intentionally a draft — I opened and developed it, so a different session reviews and merges it, not me. Nothing here needs a fix.

---

PR #230, opened 2026-09-18
review-gate: queue instead of cancel, so a comment can't erase a push's status (#92)

### LINKED ISSUE #92: review-gate: an ordinary comment can cancel the run stamping a new head, leaving it with no status
**Found on PR #167 this run, three seconds from having actually happened.** Not a defect in the marker logic — the asymmetric matching in `scripts/review-gate.mjs` is fine. It is the *delivery*: an ordinary comment can cancel the run that stamps a freshly pushed head, and its own run does no stamping, so the head can be left carrying **no `review-gate` status at all**.

`docs/ROUTINE-PROMPT.md` rule 1 is explicit about what that means: *"a head with no `review-gate` status has not been judged: check the PR's draft flag and its `REVIEW:` comments yourself before merging, never read a missing status as a pass."* The failure mode is a reviewer doing exactly that. #150 and #157 are what it costs when the gate and reality drift apart.

## The mechanism

`review-gate.yml` has one concurrency group for the whole workflow:

```yaml
concurrency:
  group: review-gate-${{ github.event.pull_request.number || github.event.issue.number }}
  cancel-in-progress: true
```

Concurrency is evaluated **at the workflow level, before the job's `if:`**. The `gate` job is deliberately skipped for a comment that does not open with a `REVIEW:`/`OWNER:` marker (that filter is what keeps the bill down — see the header comment). So a run that will do nothing still claims the group and cancels the run that would have done something.

A push to a PR followed within seconds by an ordinary comment therefore gives:

1. `synchronize` → run starts, will stamp the new head;
2. any comment → new run claims `review-gate-<n>`, **cancels (1)**;
3. that comment's `gate` job is **skipped** — it stamps nothing.

Net: the new head has no status, and nothing reports it. The workflow's own header already knows the shape of this — *"a commit status belongs to one SHA, so a push to a blocked PR produces a head carrying no `review-gate` status at all"* — and adds `synchronize` to fix it. This cancels that fix.

## Evidence, from this afternoon

PR #167, head `4c8cf32`. `GET /actions/workflows/review-gate.yml/runs`:

```
34491901881  completed  skipped    issue_comment   14:52:17Z  main               <- cancelled the one below
34491887799  completed  cancelled  pull_request    14:52:09Z  4c8cf32 (the PR)
```

The cancelled run's jobs show every step `success` — checkout and `actions/github-script` both completed — and the status landed at **14:52:16Z**, with the job killed at **14:52:19Z**. It won by three seconds. The comment that cancelled it was an ordinary progress note with no marker in it, which is why its own run is `skipped`.

That is a race, not a safety property. Push, then comment a little faster, and `4c8cf32` carries nothing.

A second symptom is visible right now on #167 even though the stamp landed: the `gate` **check run** reads `cancelled`, so GitHub reports `mergeable_state: unstable` while `/commits/<sha>/status` reads `success`. A reviewer following rule 1 gets the right answer; one glancing at the PR sees an unexplained non-green check.

## Reproduction

On any open PR: `git commit --allow-empty` is banned here, so push a real one-line change, then post a comment with no marker within a few seconds. Compare `/commits/<new head>/status` with the workflow's run list — either the `pull_request` run is `cancelled` before its `github-script` step, and the head has no `review-gate` status, or it wins the race as it did here.

## Suggested fix

The group is doing two unrelated jobs. Its stated purpose is *"two events in quick succession (comment, then Ready for review) must not race to stamp opposite states"* — that is about two runs stamping the **same** head. A `synchronize` run stamps a **different** head, so it is not in that race at all and should not share the group.

Keying the group by what is being stamped separates them:

```yaml
group: review-gate-${{ github.event.pull_request.head.sha || github.event.issue.number }}
```

Comment runs on one PR still serialise with each other; a push gets its own group and cannot be cancelled by a comment. Worth checking whether `ready_for_review` and a `REVIEW: CLEARED` comment still land in the same group under that key — they must, since that pair is the race the group exists for.

An alternative worth weighing rather than assuming: `cancel-in-progress: false` plus last-writer-wins, which trades a cancelled run for an extra billed minute. The header says a job is about a minute; the group fires rarely enough that this may simply be cheaper to reason about.

**Resolved in PR #230 using this alternative**, not the head-sha key: keying by `pull_request.head.sha` would put a `ready_for_review` event (a `pull_request` event, which carries `head.sha`) and its paired `REVIEW: CLEARED` comment (an `issue_comment` event, whose payload has no PR head sha at all) into *different* groups — breaking the exact pairing this paragraph says must hold. `cancel-in-progress: false` keeps the single existing key and avoids that trap entirely, since every run re-reads live PR state before writing regardless of group.

## The rail

Whichever fix lands, a rail should hold it: assert that `review-gate.yml`'s concurrency key is **not** the bare PR number when the workflow also carries a skip-filtered job — i.e. that a run which cannot stamp cannot cancel one that can. Text check, in `tests/unit/guardrails.test.ts` beside the existing review-gate rails.

**Added in PR #230**: asserts `cancel-in-progress: false` inside the `concurrency:` block, proved red first against the original `cancel-in-progress: true`.

## Scope

One workflow key, plus a rail and its red-proof. No `src/` change. Related to but distinct from #136 (an *edited or deleted* marker leaves a status **stale**); this leaves a head with **no** status. Neither is `review` or `debt` — a new finding, not one of the 2026-09-06 review items, and it should not be read as extending the freeze.

### Workflow (minimum scrum — the developer and the reviewer/QA must be different agents)
- [x] **Refine** — mechanism, evidence and a suggested fix are above
- [x] **Develop** — `cancel-in-progress: false` (PR #230), with the guard rail; the head-sha key considered and rejected (see above)
- [x] **Review** — a *different* agent (session_01JAmGjxnku3Kq7LuwXS3KMC): verified tests/build, mutation-tested the rail, ran the three vendored agents; found a narrower residual gap, filed as #232, not blocking
- [x] **QA** — `npm test` 1147/1147 on the merged head; the real push-then-comment reproduction is now moot given `cancel-in-progress: false` removes the mechanism entirely (no cancellation can occur on a push-vs-comment pair any more)
- [ ] **Owner action** — none
- [x] **Done** — merged `6b78ccab`, issue closed

---
_Filed by the 2026-09-10T14:37Z routine run, from its own PR #167._

---
_Migrated from #168 in the archived private repo (ugurozsahin/sky-academy-private-archive)._

ugurozsahin:
Closes #92

**Direction: tightening** — closes a gap where a review-gate check could go missing entirely (a run could then merge past no status). Nothing a run was permitted to do before is now disallowed; the workflow only got more reliable at reporting the same rule.

## What changed and why

`review-gate.yml`'s concurrency group had `cancel-in-progress: true`, but concurrency is evaluated at the
**workflow** level, before the `gate` job's own `if:` filter runs. An ordinary comment (no `REVIEW:`/`OWNER:`
marker, so its own job is always skipped) still claimed the group and cancelled a `synchronize` run seconds
from stamping a freshly pushed head. The cancelling run stamps nothing — it's skipped — so the head was left
carrying no `review-gate` status at all: exactly the gap `docs/ROUTINE-PROMPT.md` rule 1 tells a reviewer to
check for by hand ("a head with no `review-gate` status has not been judged"). Reproduced on PR #167
(evidence and full mechanism in the issue).

**Fix:** `cancel-in-progress: false`. The job always re-lists live comments and re-reads the live draft flag
right before it writes the commit status (`pulls.get()` + `listComments()`, not the webhook payload), so
nothing is lost by queuing instead of cancelling: every triggered run reaches the job, a skip-filtered comment
run finishes near-instantly without touching the status, and a run that does stamp always gets to.

**Why not the issue's primary suggestion (keying the group by `pull_request.head.sha`)?** Traced it through
and it breaks the pairing the group exists for. `ready_for_review` is a `pull_request` event — its payload
carries `head.sha`, so it would key on the head. `REVIEW: CLEARED` is an `issue_comment` event — its payload
has no PR head sha to key on at all (only `issue.pull_request.url`, no sha), so it would fall back to
`issue.number`. Those two events, which the group comment says "must not race to stamp opposite states",
would then land in **different** groups and never serialise against each other. `cancel-in-progress: false`
keeps the existing `review-gate-<pr-number-or-issue-number>` key unchanged, so that pairing still holds, and
sidesteps the sha-availability problem entirely.

- Does not touch `scripts/review-gate.mjs`'s `blockState()` logic — this is a delivery/trigger bug, not a
  predicate bug, exactly as the issue scoped it.
- Does not attempt a synthetic push-then-comment reproduction against a live PR (the issue's own suggested QA
  step) — that needs a real scratch PR and live webhook timing, which this session cannot exercise; the unit
  rail below is the reproducible half.

## Guard rail

`tests/unit/guardrails.test.ts`: "the review gate queues instead of cancelling, so a skip-filtered run cannot
cancel one that stamps (#92)" — asserts `cancel-in-progress: false` inside the workflow's `concurrency:`
block. Proved red first: added the assertion before the fix, watched it fail against the real
`cancel-in-progress: true`, then applied the fix and watched it pass.

## Tests run

| check | result |
| --- | --- |
| `npm test` | 1147/1147 (30 files, 1 new) |
| `npm run build` | clean |
| e2e | correctly out of scope (#96) — diff touches only `.github/workflows/review-gate.yml` and `tests/unit/guardrails.test.ts` |

No guard-rail budget moved (this adds a rail, doesn't change an existing budget number).

Session: https://claude.ai/code/session_017EiTZv7sPtAckGDSxSHHDU

https://claude.ai/code/session_017EiTZv7sPtAckGDSxSHHDU


ugurozsahin:
Merged `6b78ccab` (squash). Reviewed and verified on head `9be2b76` before merging:

- `npx tsc --noEmit` clean, `npm test` 1147/1147, `npm run build` clean — all match the PR's own claims.
- CI green on `9be2b76`: `test`, `branch-name`, `gate` all success; `review-gate` status success ("No open review block"); `mergeable_state: clean`; not draft; no `owner-approval` label (workflow/config change, no look change).
- Mutation-tested the new guard rail: reverted `cancel-in-progress` to `true` and re-ran the new test in isolation — 1 failed, 339 skipped (not a mistyped filter matching zero), with the exact expected/received the PR describes. Restored, re-ran the full suite clean.
- Three vendored review agents, quoted in full:
  - **pr-test-analyzer**: e2e-out-of-scope claim correct; "no budget moved" claim correct; the new rail only pins the literal YAML flag (inherent limit of testing CI YAML, not a gap specific to this PR); `tests/unit/review-gate.test.ts` correctly left untouched since this is a delivery/trigger bug, not a predicate bug. Nothing blocking.
  - **silent-failure-hunter**: found a real residual gap — GitHub Actions concurrency groups hold only one running + one pending run, so a *burst* of 3+ rapid events (not just one comment) can still silently cancel a queued marker-bearing run and leave a stale status. Reachable (this repo's own review threads show multiple comments landing within a minute), but strictly narrower than #92's original bug (which fired on any single comment) — this PR is a net improvement either way, so not blocking. Filed as #232 for a future run.
  - **type-design-analyzer**: nothing in scope — no types/interfaces/classes touched.

Ticked Review/QA/Done on #92.

Session: https://claude.ai/code/session_01JAmGjxnku3Kq7LuwXS3KMC

---

